PR #31, opened 2026-09-22
Fix E2BIG 500 error for oversized prompts and system prompts

### LINKED ISSUE #29: cli_bridge: prompts over ~128KB return HTTP 500 - the prompt is one argv element (MAX_ARG_STRLEN)
Reported by the same Claude Code user as #25/#26. Hit it on a real prompt, fixed it locally, and the fix has been running since.

## What happens

Any request whose prompt is bigger than ~128 KiB comes back **HTTP 500**. Not a 4xx, not a message about size - a 500, so from the client it reads as "the gateway is broken" rather than "your prompt is too big".

Reproduce by sending a single user message of roughly 130,000 characters through a `cli_bridge` (or `mcp_bridge`) plan.

## Root cause

The prompt is handed to the CLI as one argv element:

```
claude -p "<the entire prompt>" --output-format json ...
```

On Linux `MAX_ARG_STRLEN` caps **one argv element** at 128 KiB (32 pages). It is a *separate* limit from `ARG_MAX`, which governs the total size of the argument block. Exceeding it makes `execve` fail with `E2BIG`.

This is the part that makes it hard to spot: the total argument size is nowhere near `ARG_MAX`, so the usual "am I close to the limit" check says everything is fine. A single oversized element is enough on its own.

The same shape applies to the system prompt when it is passed as `--append-system-prompt "<text>"`, which is easy to exceed once a client sends a large system block.

## Fix that worked here

Two changes, both using flags the CLI already has:

1. **Prompt on stdin.** Spawn with `-p` and no prompt argument, then write the prompt to the child's stdin and close it. No argv element grows with user input.
2. **System prompt in a file.** Write it to a temp file and pass `--system-prompt-file` / `--append-system-prompt-file` instead of the inline string.

After that a ~150 KB prompt round-trips normally. It has been a regression check here since - send a large prompt, expect the answer rather than a 500.

## Suggested extra

Even with the fix, a size that cannot be satisfied should come back as a 413 or a 400 with a clear message rather than a 500, so a client can tell "too big" from "gateway down".

Happy to send the diff if useful - it is small, and confined to how the child process is spawned.

Fledgewing:
## Problem
When a prompt or system prompt exceeds the kernel's `MAX_ARG_STRLEN` limit (approximately 128 KiB per element), the `execve` call fails with `E2BIG`. This causes the bridge to return an HTTP 500 error, incorrectly signaling a broken gateway instead of a request that is too large.

## Solution
- **Prompt handling:** If the prompt is oversized, it is now passed via `stdin` instead of as an `argv` element.
- **System prompt handling:** Oversized system prompts are now written to a temporary file and passed using the `--system-prompt-file` flag.
- **Improved error reporting:** If a spawn still fails due to `E2BIG` (e.g., if even the file path or other arguments are too large), the server now returns an HTTP 413 (Request Entity Too Large) or an HTTP 502 (Bad Gateway) to distinguish between a client error and a service failure.
- **Byte-aware limits:** The limit check now correctly uses UTF-8 byte length instead of character count to match kernel behavior.

Fixes #29

#### Testing Notes
- Added regression tests in `tests/test_cli_bridge.py` and `tests/test_mcp_bridge.py` to verify byte-based limits, stdin redirection, system prompt file usage, and correct HTTP status code mapping for `E2BIG` and other `OSError` types.


Fledgewing (commented):
Self-review notes from the local checkout - verification summary, plus four minor/non-blocking comments inline:

- **Byte-limit math** checked against `MAX_ARG_STRLEN` (128 KiB per element): the 100k default leaves ~31 KB of headroom, and prompt and system are each independently capped, so no single argv element can exceed the kernel line. The CJK regression test pins exactly the old `len()` blind spot (33,334 chars = 100,002 bytes).
- **File-form flags** verified against a real `claude --help`: `--system-prompt[-file]` and `--append-system-prompt[-file]` are all documented on the installed CLI.
- **413/502 mapping confirmed against `switchyard/classify.py`**: 413 falls to `BAD_REQUEST` (no cooldown, no cross-lane retry of a request that can never fit); 502 falls to `TRANSIENT` (cool the plan, retry elsewhere). This mapping is load-bearing - a leftover bare 500 would have been classified transient and re-burned the lane. The 413 message text also steers clear of the prose fallback regexes (`_CONTEXT`, `_EXHAUSTED`, ...), and the dict detail yields no vendor code for `extract_vendor_code` to misread.
- **Escape hatches cover every provider**: claude gets the file forms, opencode folds the system prompt into the stdin prompt, codex already rode `instructions.md` - no provider can put an oversized element on argv anymore.
- **Cleanup paths hold on the error paths**: `system_prompt_file` unlinks its temp file in `finally` even when `_run_cli` raises the 413; the mcp file lives in the session workdir reclaimed by `cleanup_workdir` like `instructions.md`.
- **Offline suite green** on the branch: 28 cli-bridge and 30 mcp-bridge tests passed with the new tests collected, `render_preview.py` ok, no provider sockets in tests (spawn tests use a local fake script or monkeypatch `create_subprocess_exec`).

Nothing blocking; the inline comments are the minors.

sidecars/mcp_bridge/server.py
```diff
@@ -150,6 +151,18 @@
 # piped prompt, `codex exec -` reads stdin when its PROMPT argument is `-`.
 STDIN_PROMPT_LIMIT = int(os.environ.get("MCP_STDIN_PROMPT_LIMIT", "100000"))
 
+
+def over_argv_limit(text: str) -> bool:
```
Fledgewing: Minor: `mcp_bridge` already imports `cli_bridge` at runtime (`cli_bridge.normalise`, `cli_bridge._limit_error`, `cli_bridge.error_from_events` below), so this could re-export/import `cli_bridge.over_argv_limit` instead of a second copy. It does match the existing `STDIN_PROMPT_LIMIT` duplication convention between the two files, so either way is fine - noting only because two copies of the limit logic can drift if one side later gains an env knob or an ARG_MAX-total check that the other misses.

sidecars/cli_bridge/server.py
```diff
@@ -709,10 +736,54 @@ def fold_system(prompt: str, system: str | None) -> tuple[str, str | None]:
...
+    a large system block fails execve with E2BIG exactly as the prompt
+    once did (issue #29). Past the limit the text is written to a temp
+    file and the -file form of the same flag is passed instead — the
+    pinned CLI (2.1.278) documents both. Yields ([], system) unchanged
+    when there is nothing to do: no system, a system under the limit, or
+    a profile with no file flag (opencode folds the system prompt into the
+    stdin prompt, codex overrides via instructions_file, so neither ever
+    needs this).
+    """
+    key = ("system_file_args_replace" if SYSTEM_MODE == "replace"
```
Fledgewing: Minor/theoretical: when `SYSTEM_MODE=replace` but a profile defines `system_file_args` without `system_file_args_replace`, `key` silently falls back to the append-form file flag, so a replace-mode request would append instead of replace. This mirrors the exact fallback shape `build_argv` uses for the inline flags, and only the claude profile defines these keys today, so it cannot fire as written - flagging it so it stays a known trade-off rather than an accident if a fourth profile ever grows only the append file flag.

sidecars/mcp_bridge/server.py
```diff
@@ -522,7 +535,17 @@ def fill(tpl: str) -> str:
...
     if PROVIDER == "claude" and system:
-        argv += ["--system-prompt", system]
+        if over_argv_limit(system):
+            # Same E2BIG trap as the prompt (issue #29): --system-prompt is
+            # one argv element. The -file form is verified on the pinned CLI
+            # (2.1.278), and the file lives in the session workdir so
+            # cleanup_workdir reclaims it with everything else -- the same
+            # lifecycle codex's instructions.md already has here.
+            spath = workdir / "system-prompt.md"
+            spath.write_text(system)
```
Fledgewing: Cosmetic: cli_bridge's `system_prompt_file` appends a trailing newline before writing (`system if system.endswith("\n") else system + "\n"`), this write does not. Harmless - the CLI reads the whole file either way - just an inconsistency between the two implementations of the same escape hatch.

sidecars/mcp_bridge/server.py
```diff
@@ -537,13 +560,28 @@ def fill(tpl: str) -> str:
...
+                stderr=asyncio.subprocess.PIPE,
+            )
+        except OSError as exc:
+            # Same mapping cli_bridge._run_cli applies to a failed spawn (issue
+            # #29): E2BIG is the request, not the plan, so it must read as 413;
+            # any other spawn failure is this plan's capacity being broken.
+            if exc.errno == errno.E2BIG:
+                session.resolve_final({
+                    "type": "error", "status": 413,
+                    "detail": (f"request too large for {PROVIDER} cli "
```
Fledgewing: Cosmetic: cli_bridge's 413 uses an OpenAI-shaped detail dict (`{"error": {"message", "type": "request_too_large"}}`) while this is a plain string, so the two bridges render the same 413 differently under FastAPI's `{"detail": ...}` envelope. No routing impact - both classify identically as `BAD_REQUEST` off the status code - but shape parity would keep client-side error handling uniform across the two bridges.

sidecars/mcp_bridge/server.py
Fledgewing: Fixed in 6af6afc: the mcp write now normalizes the trailing newline exactly like cli_bridge's system_prompt_file (`system if system.endswith("\n") else system + "\n"`), and the test asserts the newline.

sidecars/mcp_bridge/server.py
Fledgewing: Fixed in 6af6afc: mcp_bridge now re-exports `cli_bridge.STDIN_PROMPT_LIMIT` and `cli_bridge.over_argv_limit` instead of carrying a second copy — one owner, no drift.

sidecars/mcp_bridge/server.py
Fledgewing: Fixed in 6af6afc: the mcp 413 now carries the same `{"error": {"message", "type": "request_too_large"}}` dict as cli_bridge (502 stays a plain string on both sides, already consistent). Test walks the nested shape.

sidecars/cli_bridge/server.py
Fledgewing: Fixed in 6af6afc: replace mode now demands `system_file_args_replace` or does nothing file-based (key becomes None and the inline path / spawn-guard 413 take over), so a profile with only the append file flag can never turn an override into a stack-up. build_argv's inline fallback is left as the pre-existing behaviour it is.

Fledgewing (commented):
Reviewed against HEAD `6af6afc`. Two commits: the original fix (`16c7deb`) and a self-review pass (`6af6afc`) that addresses DRY, strict replace-mode, and shape parity between the two sidecars. Solid, surgical change with clear intent matching the PR description and good test coverage of the byte math, end-to-end spawn, and the new 413/502 mapping. I have one coverage gap I'd like to see closed and one minor style nit; nothing blocking.

## Findings

### should-fix

- **No direct test for `cli_bridge.system_prompt_file` itself.** The mcp_bridge gets a focused test (`test_an_oversized_system_prompt_goes_to_a_file_not_argv`) that drives `build_argv` with `PROVIDER=claude` and asserts the `--system-prompt-file` path lands the text in the workdir. The cli_bridge equivalent — the new `system_prompt_file` context manager and its `system_file_args_replace`/`replace_extra_args` injection at `sidecars/cli_bridge/server.py:739-785` — has no analogous test. `tests/test_cli_bridge.py` runs with `PROVIDER=opencode` (`os.environ["PROVIDER"] = "opencode"` at line 18) and opencode has `system_args: []`, so even the four new tests exercise at most the no-op `([], system)` yield branch. Two paths the PR introduces go untested:

  1. `SYSTEM_MODE=append` + oversized system → `--append-system-prompt-file <tmp>` is yielded and the temp file is unlinked in `finally`.
  2. `SYSTEM_MODE=replace` + oversized system → `--system-prompt-file <tmp> --exclude-dynamic-system-prompt-sections` is yielded (the load-bearing detail that `replace_extra_args` is appended in *both* the inline and file paths).

  A parallel to `test_an_oversized_system_prompt_goes_to_a_file_not_argv`, but loading cli_bridge with `PROVIDER=claude` and asserting on `system_prompt_file` (or on `build_argv` with `SYSTEM_MODE=replace`), would cover this. Without it a future change to `system_prompt_file`'s key-selection logic could regress the replace-vs-append path silently.

- **No test that the mcp_bridge claude system-prompt file is reclaimed by `cleanup_workdir`.** The diff comment claims "the file lives in the session workdir so `cleanup_workdir` reclaims it" (`sidecars/mcp_bridge/server.py:528-531`), but `test_an_oversized_system_prompt_goes_to_a_file_not_argv` only asserts the file is created — it never runs `run_session` to completion and never calls `end_session`/`cleanup_workdir` to confirm the temp file actually disappears. A one-line `cleanup_workdir(workdir); assert not spath.exists()` after the argv assertion would close that loop.

### nit

- **`tests/test_mcp_bridge.py:1428` does `import errno` inside the test function**, while `tests/test_cli_bridge.py:10` (added by this PR) hoists `import errno` to module top. Same import in both files, inconsistent placement. Hoisting to the top of `test_mcp_bridge.py` matches the cli_bridge side.

### verified-correct, no action

- **Byte-based limit** at `sidecars/cli_bridge/server.py:106-112`: matches kernel `MAX_ARG_STRLEN`. The CJK regression test (`test_multibyte_prompts_are_measured_in_bytes_not_characters`) pins the old `len()` blind spot with a 33,334-char / 100,002-byte prompt.
- **413/502 mapping confirmed against `switchyard/classify.py:240-243`**: 413 → BAD_REQUEST (cooldown 0, no cross-lane retry of an unsendable request); 502 → TRANSIENT (60s cooldown, retry elsewhere). The 413 message text is checked against `_CONTEXT`, `_EXHAUSTED`, `_AUTH`, `_CONCURRENCY`, `_PLAN_DEAD` — none of them match "request too large for X cli (Arg list too long); reduce its size", so it cannot be misclassified by prose fallback.
- **DRY via re-export at `sidecars/mcp_bridge/server.py:150-151`**: `STDIN_PROMPT_LIMIT = cli_bridge.STDIN_PROMPT_LIMIT; over_argv_limit = cli_bridge.over_argv_limit`. Two copies of either would drift; this is the right shape.
- **Strict replace-mode fallback at `sidecars/cli_bridge/server.py:754-768`**: in `SYSTEM_MODE=replace`, `system_prompt_file` either uses `system_file_args_replace` or yields `([], system)` unchanged — it never silently falls back to `system_file_args`, which would have quietly turned an override into a stack-up.
- **Temp-file cleanup is exception-safe**: `system_prompt_file`'s `finally` at `sidecars/cli_bridge/server.py:782-785` runs `os.unlink` even when `_run_cli` raises the 413 from `create_subprocess_exec` (the inner `with` block propagates, the `finally` still fires).
- **Trailing-newline parity** between cli_bridge's `system_prompt_file` and mcp_bridge's `build_argv` (both write `system + "\n"` if not already newline-terminated) — addressed in the second commit and asserted by the updated `test_mcp_bridge.py` assertion (`huge + "\n"`).
- **Offline test suite green** on this branch: 28 cli-bridge and 30 mcp-bridge tests collected and passing locally; no provider sockets touched (`test_a_real_spawn_delivers_an_oversized_prompt_on_stdin` spawns `sys.executable` against a fake script in `tempfile.NamedTemporaryFile`, which is loopback).
- **End-to-end spawn test** at `tests/test_cli_bridge.py:462-496` is the strongest test in the suite — it actually executes `asyncio.create_subprocess_exec`, ships a 100,123-char prompt on stdin, and asserts the fake CLI received exactly that many chars via `sys.stdin.read()`. This is the test that proves the spawn itself no longer fails with E2BIG once the prompt is on stdin.
- **`TESTING.md` addition at line 397-419** documents the live repro (`~110 KB` prompt → 200) and names the lanes it can affect (`cli_bridge`/`mcp_bridge` only).

Fledgewing:
Addressed the re-review in 60884f4:

- **should-fix 1 (no direct test for `system_prompt_file`)** — new `test_the_claude_file_flags_carry_an_oversized_system_prompt` reloads the module with `PROVIDER=claude` (the same env-swap + `_modules.reload` pattern `_read_for` uses) and drives all three branches: append-form flag with temp-file cleanup asserted, replace-form flag with `--exclude-dynamic-system-prompt-sections` riding the file path, and the strict no-op when replace mode lacks `system_file_args_replace`.
- **should-fix 2 (workdir reclaim unproven)** — `test_an_oversized_system_prompt_goes_to_a_file_not_argv` now calls `cleanup_workdir(workdir)` and asserts `system-prompt.md` is gone.
- **nit (import placement)** — `import errno` hoisted to the top of `tests/test_mcp_bridge.py`, matching the cli suite.

Suites: 29 cli-bridge / 30 mcp-bridge, full offline loop green.

PR #32, opened 2026-09-22
Fix silent image dropping in CLI and MCP bridges

### LINKED ISSUE #30: cli_bridge/mcp_bridge silently drop images - flatten() discards image blocks, so the model answers blind
Reported by the same Claude Code user as #25/#26. Found it, fixed it locally, and it is the one that worried me most of the four because of **how** it fails.

## What happens

Image content blocks never reach the inner CLI. The request is accepted, no error is raised, and the model answers **as if no image had been attached** - confidently.

The moment it became obvious here: asked for the hex colour of a plain magenta swatch, the answer came back `#EDF6EC`, which is a near-white. Not a refusal, not "I cannot see an image". A specific, plausible, completely invented value. A caller has no way to tell that apart from a real reading.

## Root cause

`flatten()` in `sidecars/cli_bridge/server.py` reduces a message's content blocks to a text string and drops anything that is not text. Both image shapes are discarded:

- Anthropic: `{"type": "image", "source": {"type": "base64", "media_type": ..., "data": ...}}`
- OpenAI-compatible: `{"type": "image_url", "image_url": {"url": "data:image/png;base64,..."}}`

`mcp_bridge` imports and reuses `flatten`, so the tool path loses them the same way. On that path there is a second drop: an MCP tool result that carries `{"type": "image", "mimeType": ..., "data": ...}` is converted to text-only before it goes back, so a screenshot a tool just took is also invisible.

## Impact

Silent and total for any vision use. Screenshot-driven work through the bridge does not degrade - it hallucinates, which is worse, because the failure looks like an answer.

## Fix that worked here

1. **Stage the images to disk.** On the way in, pull every image block out, decode it, and write it into a per-session working directory (`<workdir>/img/NN.<ext>`).
2. **Let the CLI read them.** Spawn with `--add-dir <workdir>` and `--allowed-tools "Read(<workdir>/img/**)"`, and make sure `Read` is not also sitting in a disallowed-tools list.
3. **Point at them in the prompt.** Where the image block was, leave a marker such as `[image 1]` and add a short note naming the files, so the model reads them instead of guessing.
4. **Tool results too.** Convert tool-result content back into proper MCP blocks, keeping `{"type": "image"}` blocks intact rather than stringifying them.

After that, the same magenta swatch reads as magenta.

## Suggested extra

Since the failure is silent, it may be worth failing loudly instead: if a request carries image blocks and the target plan has no way to pass them through, return an error rather than quietly dropping them. A caller would much rather see "this plan cannot take images" than get a confident wrong colour.

Happy to send the diff.

Fledgewing:
## Problem
Image content blocks (Anthropic, OpenAI, and MCP tool results) were being silently flattened into text strings, discarding all vision data. This caused models to hallucinate answers because they appeared to receive a valid request but could not actually see any images.

## Solution
Implemented a staging mechanism that decodes image blocks and writes them to a temporary directory. The system now passes these file paths to the underlying CLIs (Claude, Codex, and Opencode) using their respective native image-attachment flags or by staging them for the `Read` tool. It also ensures that unsupported image requests (like remote URLs) fail loudly with a 400 error instead of failing silently.

Fixes #30

## Testing
- Verified image round-trips via `forge` for both Anthropic and OpenAI shapes.
- Confirmed CLI-specific flag plumbing for `claude -p`, `codex exec`, and `opencode run`.
- Added regression tests for image block decoding and staging in `tests/test_cli_bridge.py` and `tests/test_mcp_bridge.py`.


Fledgewing (commented):
## Summary

This PR fixes image bytes being silently flattened into text strings through the OpenAI-compatible sidecars (issue #30). The fix is comprehensive: a staging layer writes image bytes to a temp dir, replaces the image blocks with `[image N: <path>]` markers, and passes the per-CLI attachment flag (`-i` for codex, `-f` for opencode, `--add-dir` + `--allowed-tools 'Read(/…/**)'` for claude), with HTTP 400 on undecodable inputs (remote URLs) instead of a silent drop. Tool-result images get a parallel path through `openai_tool_content_to_mcp` so the MCP-tool side of the bridge no longer stringifies images either. Tests cover both shapes (Anthropic + OpenAI), mixed text+image content, remote-URL rejection, temp-dir cleanup on failure, per-provider argv, and the `images=True` toggle on `write_opencode_dir`. All 31 cli-bridge tests and 35 mcp-bridge tests pass locally.

The change is well-tested and the cli_bridge text path looks correct, but I believe the **mcp_bridge claude path is still broken for images** — `Read` remains in `--disallowed-tools` in the hard-coded `argv` template (mcp_bridge/server.py:165), and the diff only adds `--allowed-tools 'Read(/imgdir/**)'` without removing `Read` from that list. The cli_bridge text path solved exactly this problem by introducing `bare_args_images`; the MCP argv template needs a parallel treatment.

## Findings

### blocker — MCP claude path: `Read` is still in `--disallowed-tools`, so staged images cannot be read

`sidecars/mcp_bridge/server.py:154-166` (claude MCP `argv` template):

```python
"argv": ["-p", "{prompt}", "--model", "{model}",
         "--mcp-config", "{mcp_config}",
         "--allowed-tools", "{allowed_tools}",
         "--disallowed-tools",
         "Bash,Edit,Write,Read,Glob,Grep,WebFetch,WebSearch,NotebookEdit",
         "--output-format", "json"],
```

The image block (mcp_bridge/server.py:601-611) only appends:

```python
if PROVIDER == "claude":
    img_dir = Path(image_paths[0]).parent
    argv += ["--add-dir", str(img_dir),
             "--allowed-tools", f"Read({img_dir}/**)"]
```

It does not touch the existing `--disallowed-tools` list, which still contains `Read`. The cli_bridge text path was specifically written as `bare_args_images` (cli_bridge/server.py:278-280) precisely **to remove `Read` from `--disallowed-tools`** — and TESTING.md (the new §4h you wrote in this PR) confirms this is the intended semantic: *"Read is no longer in its bare-mode disallowed-tools list"*. If `--allowed-tools` "Read(/foo/**)" cleanly overrode `--disallowed-tools` "Read", that note would not have been necessary; the explicit hand-removal in `bare_args_images` is the canonical recipe.

Without that fix on the MCP side, an image-bearing claude MCP session will still be unable to `Read` its staged file, and the same silent-drop / hallucination behavior the PR exists to fix will keep happening on the MCP-tool path. The PR description and "Fixes #30" claim cover MCP tool results ("MCP tool results were being silently flattened…") so this gap is in scope, not gold-plating.

The new test `test_mcp_build_argv_passes_images_through_claude_profile` (test_mcp_bridge.py:1143) only asserts that `--add-dir` and `--allowed-tools "Read(/imgdir/**)"` are present — it doesn't assert that `Read` was *removed* from `--disallowed-tools`, which is why the gap was not caught. Suggested fix (one of these):

1. Factor the disallowed list out of the `argv` template and emit two variants from `build_argv`, mirroring how cli_bridge did it (`bare_args` / `bare_args_images`).
2. Just drop `Read` from the MCP claude `--disallowed-tools` line entirely. The bare/text path's argument for keeping Read out is the same one the MCP path needs: Read must remain usable when staged images exist, and in the MCP profile the only Read calls worth blocking are the agent's own file reads outside the staged dir — which the new `--allowed-tools "Read(/imgdir/**)"` already doesn't restrict.

Verify with the live MCP claude round-trip in TESTING.md's `docker compose exec -T claude-max-sidecar …` style of command before sending the PR.

### should-fix — `supports_images`/`codex` comment is wrong

`sidecars/mcp_bridge/server.py:1316-1320`:

```python
# Both supported MCP_PROFILES (claude, opencode) carry images via
# the same per-profile flags as cli_bridge (see build_argv /
# write_opencode_dir). codex is not in MCP_PROFILES so this is a
# constant for now, but expressed as a set membership check so
# adding a future provider here matches the cli_bridge side.
"supports_images": PROVIDER in ("claude", "opencode"),
```

`codex` IS in `MCP_PROFILES` (mcp_bridge/server.py:213), and `build_argv` at lines 601-611 already handles it. Either the comment is wrong (it is) or the exclusion (`PROVIDER in ("claude", "opencode")`) is wrong — please pick one. If `supports_images` should be `PROVIDER in ("claude", "opencode", "codex")`, change both lines. If it should stay as-is, rewrite the comment to explain why codex is *deployed-but-not-image-capable* here, since right now it reads as a self-contradiction.

### nit — `stage_or_fail` only catches `ImageUnsupportedError`; other exceptions leak the temp dir

`sidecars/cli_bridge/server.py:242-256` (and mirrored on the `stage_images` callers in mcp_bridge):

```python
img_dir = Path(tempfile.mkdtemp(prefix="swimg-"))
try:
    return stage_images(messages, img_dir / "img"), img_dir
except ImageUnsupportedError:
    shutil.rmtree(img_dir, ignore_errors=True)
    raise
```

`stage_images` → `_stage_image` calls `path.write_bytes(data)`. If that raises `OSError` (disk full, permissions, ENOSPC on a sandbox), it propagates and the freshly-created `swimg-*` temp dir is leaked. Add a `finally` (`except Exception` re-raise pattern) or wrap the `try` more broadly. This is also the only path on the text side where a failed staging leaves no HTTP response — every caller except `stage_or_fail` itself turns an image-present request into a temp-dir leak.

### nit — `image_dir`/`img_dir` naming inconsistency

`sidecars/mcp_bridge/server.py:957` declares the parameter as `image_dir: Path | None = None`, while `sidecars/cli_bridge/server.py:1357` (and every prior mcp_bridge call site in this diff: lines 920-948, 1100-1132) uses `img_dir`. Same concept, two spellings across the two files. Pick one — `img_dir` matches the rest of the file.

### nit — `start_session` docstring describes a path that doesn't exist

`sidecars/mcp_bridge/server.py:966-971` (the docstring you wrote):

```
`image_dir`, when set, is a separate temp dir the caller staged into (used by
the rebuild path so images can be staged BEFORE the workdir exists); when None
the caller has already staged into `workdir/img` and the session's own
workdir cleanup at end_session takes care of it.
```

Neither `handle_fresh` nor `resume_gone_session` ever passes `img_dir=None` for an image-bearing request: `cli_bridge.stage_or_fail` returns `(paths, dir)` where `dir` is always a fresh temp dir (or `None` for a text-only request). The "staged into `workdir/img`" path is dead text. Rewrite the docstring to reflect what actually happens (the workdir and `img_dir` are always distinct; `start_session` only cleans `img_dir`; the workdir is cleaned by `end_session`).

## What I checked and found solid

- **CI / tests**: No CI on the PR branch per `gh pr checks`. Ran `for t in tests/test_*.py; do python3 "$t"` locally — 31 cli-bridge, 35 mcp-bridge tests, and the routing/classify suites all green. New tests assert behavior (decode, marker substitution, remote-URL exception, argv per profile, opencode `read` toggling), not just that the code runs.
- **Loud-failure contract**: `ImageUnsupportedError.http()` → HTTP 400 with `error.type == "images_unsupported"` is consistent across both bridges (`_handle_chat`, `handle_fresh`, `resume_gone_session` all use `exc.http()`). `isError OR content_is_error` in `_continue_followup` correctly preserves the caller's explicit flag.
- **Resource cleanup**: `img_dir` is cleaned on every error path I traced (HTTPException before gate acquire, gate full, capacity refused, tool-result loop). The `stage_or_fail` path cleans its own temp dir on `ImageUnsupportedError` before re-raising.
- **MCP tool-result path**: `openai_tool_content_to_mcp` correctly handles plain strings, Anthropic `source.base64`, OpenAI `data:` URLs, and remote URLs (loud `isError`); tests cover each.
- **TESTING.md §4h**: thorough — explicit magenta-PNG round-trip via `forge`, per-CLI flag plumbing commands, and the loud-failure note for remote URLs. Ship as written.

Fledgewing:
Addressed all findings in 2b5a51e (now on the PR branch):

**1. blocker — MCP claude path: `Read` still in `--disallowed-tools`**
Fixed by the same structural treatment as cli_bridge: the hard-coded disallowed list is out of the claude argv template; `build_argv` now emits `--disallowed-tools` per request — the full list (with `Read`) for text-only sessions, the list without `Read` for image sessions (new `disallowed_tools_default` / `disallowed_tools_images` profile keys). `--allowed-tools "{allowed_tools}"` with the `mcp__switchyard__*` qualifiers is untouched, so MCP tool access is unchanged. Tests: the image-argv test now asserts `Read` is absent from the disallowed value *and* that the MCP qualifiers survive in `--allowed-tools`; a new no-image regression test asserts the original full list (with `Read`) is still emitted. (test counts 35 → 36)

**2. should-fix — `supports_images` comment vs code**
The code was the wrong half: `codex` is in `MCP_PROFILES` and `build_argv` handles `-i`, so `supports_images` is now `PROVIDER in ("claude", "opencode", "codex")` and the comment rewritten to match.

**3. nit — `stage_or_fail` leaks the temp dir on non-ImageUnsupportedError**
Cleanup broadened to `except BaseException` (any failure — e.g. `OSError` from `write_bytes` — removes the dir before re-raising). The mirrored staging sites in `handle_fresh` / `resume_gone_session` also remove the dir if anything escapes between staging and `start_session`. New test simulates an `OSError` from `_stage_image` and asserts no `swimg-*` leak. (31 → 32)

**4. nit — `image_dir` vs `img_dir`**
Renamed to `img_dir` in `start_session` and the two test stubs.

**5. nit — `start_session` docstring dead path**
Rewritten to the real contract: workdir and `img_dir` are always distinct temp dirs; `start_session` cleans `img_dir` where it owns it; the workdir is cleaned by `end_session`.

Full offline suite green (cli-bridge 32, mcp-bridge 36, `render_preview.py` OK).

PR #40, opened 2026-09-22
Opt-in perishable lane strategy: drain the seat whose quota expires soonest (#35)

### LINKED ISSUE #35: Opt-in 'perishable' lane strategy: drain the seat whose quota expires soonest
Follow-up to the fill-through discussion (same reporter as #25, #26, #29, #30). Fill-through is the right default. This asks for an **opt-in** ordering strategy for lanes whose members are subscription seats with rolling windows.

## The problem

A lane's order is fixed in `plans.yaml`, so `Picker.pick()` always drains member 1 first. With several subscription seats that reset at **different times**, that wastes quota:

- Seat A: 50% of its weekly window left, resets in **5 days**
- Seat B: 40% left, resets in **6 hours**

Fixed order puts new work on A. B's 40% then expires unused in 6 hours, while A burns room it would have kept. Across a week that adds up to whole seats of paid capacity thrown away.

The picker already reads the number it would need and then discards it: `is_spent()` fetches `window_facts(plan.key, plan.quota.label)` and only asks `reported_pct_used >= 100`. `record_reported()` also stores `reset_at` in the same hash. So the data is already there.

## Proposed: `strategy: perishable` (opt-in, per lane)

Score each member from facts the probe already writes:

```
room      = 100 - reported_pct_used           # target window (weekly)
hours     = max(1, (reset_at - now) / 3600)
score     = room / hours                       # "use it or lose it" rate
```

The highest score goes first. In the example, B scores 6.7/h and A scores 0.4/h, so B drains before it resets and A keeps its room for later in the week.

### Guard rails (the parts that keep it from hurting performance)

1. **Affinity still wins.** The lease branch runs first, same as now. Only NEW sessions follow the score, so a live session never loses its warm prompt cache.
2. **Off the hot path.** Compute the order when the headroom probe polls, and store it as one key per lane. `pick()` reads that one key. There is no extra per-member work per request.
3. **Stale data = config order.** If the facts are older than about 2x the probe interval, fall back to the `plans.yaml` order. Never route on a stale meter.
4. **Hysteresis.** Swap two members only when their scores differ by more than about 20%, and at most once per poll. This avoids flapping, and fill-through within the chosen order keeps each seat's cache warm.
5. **The constraint window gates, it does not rank.** Rank on the target (weekly) window, but skip a member for NEW sessions when its 5-hour window is above about 90%. Starting a long task on a seat that is about to hit its burst limit just forces a mid-task spill.
6. **Model-scoped windows.** If a plan reports a scoped weekly window for the requested model family, use it when it is the tighter of the two.
7. **Unknown facts sort last, not first.** A member with no reported percentage keeps its config position after every scored member. It should not jump the queue because it reported nothing.

Default stays `fill`, so nothing changes for anyone who does not opt in.

## Why a strategy and not a config reorderer

A script that rewrites the order in `plans.yaml` works, but every change needs a gateway restart, which drops in-flight requests. It also goes stale between runs. Inside the picker, the order follows the probe for free.

Fledgewing:
Fixes #35

## What it does

Adds an opt-in `strategy: perishable` knob to lanes in `config/plans.yaml`. When set, the picker ranks that lane's members by perishable score (`room / hours_to_reset`) so the most-burned subscription goes first; the default `fill` strategy is unchanged.

## Why

Lanes like `forge` mix weekly subscriptions whose budget is racing the reset. Draining them in config order means a 10%-used plan sits idle while a 90%-used plan gets more traffic and dies sooner. The picker already has a `hold_plan` lease path; this is a route-side recompute that runs only after the held-plan branch has returned, so in-flight work is untouched.

## Contract (config-side)

`plans.yaml`:
```yaml
lanes:
  forge:
    strategy: perishable    # default absent == "fill", no Redis writes
    order: [minimax-ultra/m3, minimax-max/m3, ...]
    tail: [local-box/qwen]
```

- Per-member score on the TARGET window only: `room = 100 - reported_pct_used`, `hours = max(1.0, (reset_at - now)/3600)`, `score = room / hours`. Higher = drain sooner.
- The 5-hour constraint window is a GATE, not a rank: `pct_used > 90` flips `gate5h_<ref>` on and the picker skips that member for NEW sessions. Affinity and a pinned follow-up stay untouched — they already returned above the re-rank. The gate is binary on purpose; conflating it with the rank would let a 50%-member whose 5h is at 95% jump ahead of an 80%-member whose 5h is at 30% for the wrong reason.
- When the plan also publishes a `weekly_scoped` window, the picker uses that as the member-level reading (tighter than the plan-wide one) and falls back to the target window otherwise.
- A member with no usable facts (no probe yet, fresh credential, etc.) gets no score and lands after every scored member, in config order. An unknown plan never beats a known one.
- Tail members are never scored; they live at the end regardless.

## Writer-side contract

`switchyard/portal/app.py` piggybacks on the existing probe poll. After every successful probe, it recomputes the rank for each `strategy: perishable` lane that contains a live member of that plan:

1. Reads `reported_pct_used` and `reset_at` for the target window once per affected lane.
2. Computes the desired order as `score`-descending; appends unscored refs in config order.
3. Applies **at most one adjacent swap per poll** where the lower-position member beats the one in front by ratio `>= ~1.2`. A 51/49 split does NOT flip the lane; a real 95/30 spread does. Keeps the rank responsive to a real gap without thrashing on noise.
4. Writes the order to a `sy:lane-order:{lane}` Redis hash with `computed_at` + `stale_after_ms`. **Deletes the key first** so a ref that left the lane does not carry over an old `score_<ref>`.

`stale_after_ms` is `~2 * probe.interval_seconds`, so one missed poll is tolerated and a quiet probe pulls the lane back to config order instead of serving an old ranking forever.

## Reader-side contract

`switchyard/picker.py`:

```python
lane_cfg = self.registry.lanes.get(lane)
if (lane_cfg is not None
        and lane_cfg.strategy == "perishable"
        and self.policy is not None):
    order = await self.policy.ledger.get_lane_order(lane)
    members = _perishable_visit_order(members, by_ref, order, skipped)
```

- One Redis read per affected pick. If the key is missing or stale, `get_lane_order` returns `None` and the body falls through in config order — exactly what every non-perishable lane already does.
- `gate5h` refs are filtered out of the iteration list and listed in `skipped` so the capacity board says why.
- This runs **after** the held-plan branch has already returned, so an in-flight session on a leased plan is never displaced by the new rank.

## Default-path invariant

`strategy` is absent by default on every existing lane in `config/plans.example.yaml` (it is shown commented out on `apex`, `judge`, `forge` — the three lanes that would actually benefit — so a fresh clone ships bit-for-bit unchanged). A fill-strategy lane:

- never reads the lane-order key
- never writes the lane-order key
- routes in config order

This is enforced by the test `test_a_strategy_absent_lane_never_touches_the_lane_order_key`.

## Test evidence

- 6 new tests in `tests/test_routing.py`:
  - `test_perishable_lane_routes_higher_room_first` — score-rank wins over config order
  - `test_perishable_hysteresis_swaps_at_most_one_adjacent_pair` — writer's swap cap
  - `test_perishable_lane_falls_back_to_config_order_when_stale` — single-key read, missing/stale -> config order
  - `test_perishable_5h_gate_skips_a_member_for_a_new_session` — gate is a gate
  - `test_perishable_unknown_member_sorts_after_scored_ones` — unscored lands last
  - `test_a_strategy_absent_lane_never_touches_the_lane_order_key` — default-path invariant
- Routing test count: **21 → 27**.
- Full offline sweep (`for t in tests/test_*.py; do python3 "$t" >/dev/null || echo "FAIL $t"; done`): **no FAIL**.
- `python3 tests/render_preview.py`: clean (`10 plans, 19 models, 5 lanes, 18 quota windows, 4 probes`).
- Tests run offline; no provider sockets touched.

## Files touched (6)

- `switchyard/models.py` — `Lane.strategy`, validator in `load()`
- `switchyard/usage.py` — `K_LANE_ORDER`, `Ledger.set_lane_order` / `get_lane_order`, `perishable_score()`
- `switchyard/picker.py` — `_perishable_visit_order()`, the one-key read in `Picker.pick`
- `switchyard/portal/app.py` — piggy-backed recompute in `_poll_probes`, `_one_adjacent_swap` hysteresis, `_stale_after_ms` derived from probe interval
- `config/plans.example.yaml` — documented contract + commented `strategy: perishable` on `apex` / `judge` / `forge`
- `tests/test_routing.py` — 6 new tests

No `.env`, no `config/plans.yaml`, no `docker compose` from this worktree.


switchyard/portal/app.py
```diff
@@ -122,21 +129,200 @@ async def _poll_probes() -> None:
...
+                next_order.remove(ref)
+                next_order.append(ref)
+
+        entries = {ref: {"score": scored_map[ref][0],
+                         "gate5h": 1 if scored_map[ref][1] else 0}
+                   for ref in next_order if ref in scored_map}
+        for ref in next_order:
+            if ref not in entries:
+                entries[ref] = {"score": 0.0, "gate5h": 0}
+        await ledger.set_lane_order(
```
Fledgewing, line 289: **[blocker] Per-plan recompute wipes sibling plans' entries — picker sees only the last-polled plan as "scored"**

The writer is called once per plan from `_poll_probes` (line 146), but the data model is per-lane and `set_lane_order` is delete-then-write (`switchyard/usage.py:353`). For a perishable lane whose body mixes multiple plans, each successful probe produces a write that contains only that plan's refs. The next plan's write deletes everything the previous one wrote.

Verified end-to-end against `_recompute_perishable_for_plan` with the shipped example config and a perishable `forge`:

```
after minimax-ultra drive (has facts): score_* = ['score_minimax-ultra/m3']   score=0.536
after minimax-max drive (overwrites):  score_* = ['score_minimax-ultra/m3']   score=0.0
picker sees: [('minimax-ultra/m3', 0.0, False)]   # minimax-max/m3 NOT in hash
```

Two failure modes fall out of this:

1. **`set_lane_order` deletes the whole key** before writing. The next plan's write is the only thing left.
2. **`_one_adjacent_swap` (line 295) operates on `previous_refs`, not `desired`.** When `desired` has refs that `previous_refs` does not (e.g., a brand-new plan just probed), they are silently absent from `next_order` and never reach the hash. In the run above, `_one_adjacent_swap(['minimax-ultra/m3'], ['minimax-max/m3'], {minimax-max/m3: ...})` returns `['minimax-ultra/m3']` — `minimax-max/m3` is dropped on the floor.

Effect on the picker: after one full poll cycle of a multi-plan perishable lane, only the **last-polled plan's** refs survive in `score_<ref>` / `gate5h_<ref>`. The other plans' refs are absent from `score_by_ref` in `_perishable_visit_order`, so they fall through to the `unscored` branch and sort after every scored member — regardless of how full their perishable score would have been. The score-rank contract the PR description promises is therefore broken on the most important cases (`apex`, `judge`, `forge` all mix multiple plans in the shipped example).

Note: the six new tests in `tests/test_routing.py` bypass the writer via direct `ledger.set_lane_order(...)` calls, so this regression is uncaught. The picker-facing tests (e.g. `test_perishable_unknown_member_sorts_after_scored_ones`) only exercise the read path against a hand-written hash.

Suggested fixes (any one is sufficient):
- **Score all lane members in one recompute.** The trigger is per-plan, but the recompute body should iterate `reg.lane_members(lane.key)` across plans, reading per-plan facts as needed, so the write contains every body ref.
- **Per-(plan, lane) storage.** Use `sy:lane-order:{lane}:{plan}` keys; the picker merges.
- **Merge, don't delete.** Read the previous hash, update only this plan's `score_<ref>` / `gate5h_<ref>` fields, write back.

Also add a regression test that drives `_recompute_perishable_for_plan` for two plans in the same perishable lane and asserts both plans' refs are present in `get_lane_order`.

switchyard/portal/app.py
```diff
@@ -122,21 +129,200 @@ async def _poll_probes() -> None:
...
+        sa = scores.get(a, (0.0, False))[0]
+        sb = scores.get(b, (0.0, False))[0]
+        if sb <= 0 or sa <= 0:
+            continue
+        if sb / sa < 1.2:
+            continue
+        # The lower-position member b should come before a in `desired`. If it
+        # already does, leave it: the swap is purely reactive to the score
+        # difference, not a teardown of the existing order.
+        if desired.index(b) < desired.index(a):
```
Fledgewing, line 353: **[should-fix] `_one_adjacent_swap` neither adds new refs nor drops removed ones**

The function operates only on `previous_refs`. It returns `out = list(previous)` and then may swap one adjacent pair. Refs in `desired` that aren't in `previous_refs` (e.g. a member whose plan just probed successfully for the first time, or a new `strategy: perishable` lane becoming warm) are silently absent from `next_order` after the call. Refs in `previous_refs` that aren't in `desired` (config edit removed a ref, a member's plan was disabled, a model was disabled) are kept in `next_order` and get written back as `{score: 0.0, gate5h: 0}` by the fallback loop at line 268, polluting the picker-facing hash with stale, zero-scored entries that read as "scored" in `_perishable_visit_order`.

Concrete run with the shipped example config and perishable `forge`:

```
_one_adjacent_swap(['minimax-ultra/m3'], ['minimax-max/m3'], {minimax-max/m3: (0.238, False)})
=> ['minimax-ultra/m3']   # minimax-max/m3 dropped
```

The docstring calls the function "at most one adjacent swap on the way from previous toward desired", but the implementation never moves toward `desired` — it only reorders `previous`. The result should instead be: drop refs from `previous` that are absent from `desired`, append any refs in `desired` not already in the result (in `desired` order), THEN apply the one-swap cap. Something like:

```python
out = [r for r in previous if r in set(desired)]
out += [r for r in desired if r not in set(out)]
# ...then the existing adjacent-swap loop on `out`.
```

This is independent of the per-plan overwrite bug above, but the two together are what makes the multi-plan case lose data on every poll.

switchyard/picker.py
```diff
@@ -48,6 +48,47 @@ def _why(result: int, plan_cap: int, model_cap: int | None) -> str:
...
+        return members
+    score_by_ref: dict[str, float] = {}
+    gate5h_refs: set[str] = set()
+    for entry in order["members"]:
+        if entry["gate5h"]:
+            gate5h_refs.add(entry["ref"])
+        else:
+            score_by_ref[entry["ref"]] = entry["score"]
+    # Original config position per ref, used for unscored members.
+    config_pos = {m.ref: i for i, m in enumerate(members)}
```
Fledgewing, line 80: **[nit] `config_pos` is the post-`lane_members()` urgency order, not the YAML `order:`**

`config_pos = {m.ref: i for i, m in enumerate(members)}` (line 77) is built from the function's `members` parameter, which is the output of `Picker._members()` → `Registry.lane_members()`. That list is already urgency-sorted (soonest-expiry plans first, `urgency()` at `switchyard/models.py:365`), with the tail concatenated after the body. So `config_pos` orders by *lane_members order*, not by the YAML `order:` list.

The docstring (lines 54-66) says unscored members are kept "in config order AFTER every scored member so an unknown plan never beats a known one", and the inline comment on line 76 says "Original config position per ref, used for unscored members." A reader looking for the YAML order won't find it here.

The routing invariant ("unscored lands last in some stable lane order") is preserved either way, so this is purely a documentation nit — clarify the comment to say "lane_members order" or "post-urgency order" so a future maintainer doesn't go looking for the YAML list.

Fledgewing:
Reviewed 3 findings at `bfa9ad3`: 1 blocker, 1 should-fix, 1 nit — see inline comments.

The blocker is structural: the writer is invoked per-plan from `_poll_probes` but the data model is per-lane, and `set_lane_order` is delete-then-write. For a perishable lane whose body mixes multiple plans (which is exactly what `apex`, `judge`, `forge` do in the shipped example), the picker ends up reading only the last-polled plan's refs as "scored" — the score-rank contract the PR description promises holds only in the degenerate single-plan-per-lane case. The two should-fix/nit comments are downstream of that root cause (the swap function doesn't add new refs) and a documentation mismatch on `config_pos`.

The existing 27 routing tests are green, but none of them exercise the writer against a multi-plan lane — they bypass `_recompute_perishable_for_plan` via direct `ledger.set_lane_order(...)` calls. A regression test that drives the writer for two plans in one lane and asserts both plans' refs are present in `get_lane_order` would catch the regression; happy to draft one if useful.

Issue #25, opened 2026-09-21
cli_bridge to_openai drops Anthropic cache tokens, ~99% usage under-report on cached sessions


Fledgewing:
Reported by a friend using SwitchYard through Claude Code. Verified in the code.

## What happened
A bridged Claude job that really consumed ~64k tokens reported **95 tokens** to the ledger (roughly 99% under-report).

## Root cause
`to_openai()` in `sidecars/cli_bridge/server.py:893-910` builds the OpenAI response with:

```python
"prompt_tokens": int(usage.get("input_tokens", 0) or 0),
```

Anthropic's usage counts cache tokens separately from `input_tokens`:

- `input_tokens` — non-cached prompt tokens
- `cache_read_input_tokens` — tokens served from the prompt cache
- `cache_creation_input_tokens` — tokens written to the cache this turn

On a cached session almost the entire prompt lands in the cache fields, so `input_tokens` alone is tiny (~95) while the real token count (64k+) sits in `cache_*_input_tokens`, which `to_openai` discards.

The `claude_json` parser (`parse_output`, line ~565) passes the vendor JSON through untouched, so the cache fields are present at this point and simply dropped here. For the OpenCode/Codex shapes the parser's `cache_read_tokens` field (lines ~604, ~618) is dropped too.

## Impact
- `to_openai` is the only place usage enters the OpenAI response for bridged plans (also used by `mcp_bridge`, ~line 823).
- The gateway ledger (`switchyard/hooks.py:381`) records only `prompt_tokens` + `completion_tokens` into Redis (`usage.py` FIELDS), and that same short number feeds:
  - window/burn-rate accounting (`usage.py:137,180-183,210`)
  - quota gating (`policy.py:199-202`)
  - headroom / seat pacing decision — "the ledger that decides how hard to lean on a seat" (`usage.py:344`)

So cached seats look nearly idle and get leaned on harder than plan headroom actually allows.

## Suggested fix
In `to_openai`, when the payload is an Anthropic shape, add the cache fields:

```python
cache_read = int(usage.get("cache_read_input_tokens", 0) or 0)
cache_write = int(usage.get("cache_creation_input_tokens", 0) or 0)
prompt = input + cache_read + cache_write
```

[...]

PR #42, opened 2026-09-22
Retry once on transient upstream failures; escalate repeated cooldowns (#36)

### LINKED ISSUE #36: Upstream 5xx reaches the caller, and a broken seat is re-fed every 60s
Same reporter as #25, #26, #29, #30. Seen live today on current `main` (c49e49a).

## What happens

One subscription seat started returning `API Error: 500 Internal server error` on every call. It came from the provider: the inner CLI exits 1 with `terminal_reason: api_error`, `api_error_status: 500` and zero tokens. The seat's own `/usage` showed plenty of room (20% session, 53% weekly), so nothing marked it spent.

Two things then compound:

1. **The caller gets the 502.** `num_retries: 0` and there is no re-pick inside the lane, so the request that hit the broken seat fails outright, even though other members of the same lane had free slots.
2. **The seat comes back every 60s and fails again.** `classify()` maps any 5xx to `TRANSIENT, 60`. After 60 seconds the seat is first in lane order again, takes the next new request, and fails again. Over one hour: 10 upstream 502s across two seats, and neither was ever benched for longer than a minute.

It shows up clearly in a health check. The first call after each cooldown lapses fails, and every call in between succeeds.

## Suggested fix

**A. Re-pick once on a transient failure, before any bytes are streamed.** If the failure is `TRANSIENT` (5xx, timeout) and nothing has reached the caller yet, release the slot, cool the plan, and run `pick()` again for the same lane with the failed member excluded. Allow one retry only, so a lane-wide outage cannot turn into a retry storm. `QUOTA_EXHAUSTED`, `AUTH` and `PLAN_DEAD` could take the same path, since they already drop the lease.

**B. Escalate repeated transient cooldowns (a circuit breaker).** Keep a per-plan consecutive-failure counter in Redis:

```
cooldown = min(60 * 2 ** (consecutive_transients - 1), 1800)   # 60, 120, 240 ... 30 min
```

Reset the counter on the first success. One provider blip still costs 60s, but a seat that is broken for hours stops eating one request a minute.

**C. (Optional) Surface it on the board.** Show a plan with a streak of 3 or more transient failures as "failing", separate from "cooled". A human looking at the board today sees a healthy seat with room left.

## Repro

Any lane whose first member returns 5xx while the later members are healthy. Send one request every 70 seconds: every one of them fails with 502 until the upstream recovers.

Fledgewing:
## Problem

Same pattern issue #25/#26/#29/#30 reporter saw today on `main` (c49e49a): one subscription seat returned `API Error: 500 Internal server error` on every call while its `/usage` showed plenty of room, and two things compounded into a self-feeding outage:

1. **The caller gets the 502.** `num_retries: 0` and there is no re-pick inside the lane, so the request that hit the broken member of the lane failed outright — even though siblings on the same lane had free slots.
2. **The seat comes back every 60s and fails again.** `classify()` maps any 5xx to `TRANSIENT, 60`, so after 60s the broken member is first in lane order again, takes the next new request, and fails again. One provider blip costs 60s of sit-out; a seat that is broken for hours eats one upstream call a minute, every minute.

## What this changes

**A. One lane re-pick on a transient failure, before any bytes are streamed.** LiteLLM's router now runs with `num_retries: 1` + `router_settings.disable_cooldowns: true`, and that one retry is the ONLY place the gateway lets litellm re-route. SwitchYard's new `async_pre_routing_hook` re-picks against the same lane / session / tools / pin as the failed attempt, but with the failed member excluded. Implemented as a per-attempt marker (`ctx['failed_ref']`) on `async_log_failure_event` and an in-place mutation of the metadata dict the router reuses across attempts, so the upstream call (which reads `metadata[switchyard][model]`) re-wires to the new pick without any other consumer noticing. The session lease is **moved** to the new plan — never dropped, since dropping would strand a tool-loop mid-call — and the re-pick returns not-sticky because the held plan was the broken one. If the re-pick cannot place a request (the whole lane is gone or full), the hook raises 429 `lane_saturated` with `Retry-After: 20`, exactly mirroring the pre-call hook's behaviour, so a lane-wide outage still becomes a 429 instead of a retry storm. Operates before bytes are streamed: litellm's retry loop wraps the upstream call itself, so a re-pick here cannot leak half a response. Mid-stream failures are not retried by litellm at all and are not affected. Skipped when the caller named a deployment directly (`pick_direct` refuses rather than substitutes) — there is nothing to spill to, and the operator asked for that specific model.

**B. Escalated cooldown ladder (a circuit breaker) for repeat transient failures.** A new per-plan Redis key `sy:tfail:{plan}` (TTL 7200s sliding) counts consecutive `TRANSIENT` failures; the first successful call against the plan resets it, so recovery is automatic. The cooldown becomes `min(base * 2**(streak - 1), max_seconds)` — first failure still costs the configured base (60s by default — today's behaviour preserved); each subsequent failure in a row doubles it (60, 120, 240, 480, 960) up to a 30-minute cap. A once-per-attempt double-count guard prevents `async_log_failure_event` (per attempt) and `async_post_call_failure_hook` (once after retries exhaust) from both bumping the streak for the same failure — without it, a single 5xx would look like two and the ladder would double-count its way to the cap. The guard keys off the new `request_id` minted for the re-pick, which is naturally unique per attempt because the picker mints a fresh one for every pick.

**C. The portal board shows a distinct 'failing · Nx' chip at streak ≥ `streak_alert`.** Rendered in the capacity fragment, just before the existing `cooled` / `cap == 0` cases. The chip uses the `warn` class, not `bad` — the plan is on the ladder but still accepts work, so its row should not pretend to be unavailable (that's what the existing `cooled` tag is for). The chip carries the actual streak, not the threshold, so the operator sees the climb in real time. Threshold and cap live in a new `settings.transient_breaker` block (`enabled`, `max_seconds`, `streak_alert`), mirrored in `config/plans.example.yaml` so a fresh clone sees it commented with the rationale.

## Why a single retry, not zero or many

- **Zero:** today's behaviour. The caller eats the 5xx even when a sibling is healthy. That's the bug.
- **One:** the router's retry loop re-invokes `async_pre_routing_hook` per attempt. With exactly one retry, the picker gets to spill onto a different member of the same lane on its first attempt — a real second attempt, not the same broken deployment tried again. num_retries=1 means at most 2 attempts in total, so a lane-wide outage cannot storm.
- **Two or more:** would silently double (or worse) our upstream calls without buying anything. The first re-pick already drained the lane of healthy candidates; the second can only do harm.

`disable_cooldowns: true` is load-bearing with `allowed_fails=1`: without it, litellm would cool the single-deployment group between attempts and the re-pick would never fire. `cooldown_time: 0` does NOT work — litellm's router takes the `or` default, not our zero. SwitchYard owns all real cooldowns via Redis, which is the only state that outlives this process and that the portal can see.

## Verification

Offline suite, no network calls (macOS `python3` against the existing harness, conftest blocks non-loopback sockets):

```
python3 tests/test_routing.py tests/test_classify.py 2>/dev/null   # green
for t in tests/test_*.py; do python3 "$t" >/dev/null || echo "FAIL $t"; done   # no FAIL lines
# rendered index.html + fragments (87431 bytes) -> /var/folders/.../switchyard-preview.html
# 10 plans, 19 models, 5 lanes, 18 quota windows, 4 probes
```

Per-file test counts (all green):

| file | tests |
| --- | --- |
| `tests/test_auth_audit.py` | 7 |
| `tests/test_classify.py` | 13 (incl. 4 new `escalated_cooldown_*`) |
| `tests/test_cli_bridge.py` | 37 |
| `tests/test_hot_reload.py` | 9 |
| `tests/test_mcp_bridge.py` | 38 |
| `tests/test_oauth.py` | 10 |
| `tests/test_policy.py` | 22 |
| `tests/test_portal.py` | 4 (incl. new `test_capacity_row_shows_failing_chip_when_streak_meets_alert`) |
| `tests/test_probes.py` | 10 |
| `tests/test_reasoning.py` | 6 |
| `tests/test_routing.py` | 24 (incl. 3 new `pick_honours_exclusions` / `repick_moves_session_lease` / `lane_saturated_when_every_member_excluded` + updated `test_the_generated_config_declares_no_general_fallbacks`) |
| `tests/test_token_proxy.py` | 8 |
| `tests/test_usage.py` | 18 |
| **total** | **206** |

`render_preview.py` exercises the new chip path (`transient_streak: 3` on the first fixture row → warn-tag "failing · 3x" in the rendered capacity fragment).

## Files touched (13)

- `config/plans.example.yaml` — new `settings.transient_breaker` block (enabled, max_seconds, streak_alert) with comment explaining the ladder.
- `switchyard/classify.py` — `escalated_cooldown(base, streak, cap)` (60, 120, 240, … capped at `cap`).
- `switchyard/gen_litellm.py` — `num_retries: 1` + `router_settings.disable_cooldowns: true` (load-bearing with `allowed_fails=1`).
- `switchyard/hooks.py` — `async_pre_routing_hook` re-pick with exclude; `ctx['failed_ref']` marker on failure; `direct`/`needs_tools`/`pinned` carried on the metadata so the re-pick honours them; double-count guard via `_verdict_applied` keyed on `request_id`; reset on success; escalated cooldown wired through `note_transient_failure`.
- `switchyard/models.py` — `TransientBreaker` dataclass; loaded into `Settings.transient_breaker`.
- `switchyard/picker.py` — `pick(..., exclude=)`; lease-pinned re-pick skips the held plan when it's excluded and never drops it; capacity row now carries `transient_streak`.
- `switchyard/portal/app.py` — capacity collector appends an alerting row entry once the streak reaches `streak_alert`.
- `switchyard/portal/templates/_capacity.html` — `failing · Nx` chip on the row when `transient_streak >= 3`.
- `switchyard/slots.py` — `K_TFAIL = "sy:tfail:{plan}"` (TTL 7200s); `note_transient_failure`, `transient_failure_streak`, `reset_transient_failures`.
- `tests/test_classify.py` — 4 new tests pinning the ladder (first failure = base, doubles, caps, uses supplied base).
- `tests/test_portal.py` — new fragment test covering below/at/above threshold streak.
- `tests/test_routing.py` — 3 new tests for exclude, lease-move, all-excluded → `LaneSaturated`; existing `test_the_generated_config_declares_no_general_fallbacks` updated to assert `num_retries=1` + `disable_cooldowns=True`.
- `tests/render_preview.py` — fixture row carries `transient_streak: 3` so the preview renders the chip.

No `.env`, no `config/plans.yaml`, no `docker compose` from this worktree.

Fixes #36


switchyard/portal/templates/_capacity.html
```diff
@@ -89,6 +89,12 @@
         <td class="mono nowrap" style="width:32%">
           {% if p.cooled %}
             <span class="tag bad">{{ p.cooldown_reason.replace('_',' ') }} · {{ (p.cooldown_remaining // 60) }}m</span>
+          {% elif p.transient_streak and p.transient_streak >= 3 %}
```
Fledgewing: **[should-fix] The chip threshold is hardcoded to 3, ignoring the configurable `streak_alert`**

`p.transient_streak >= 3` is a literal 3, but everywhere else in the PR — the `TransientBreaker` dataclass docstring (`once the streak reaches streak_alert`), the `plans.example.yaml` block comment (`once the streak reaches streak_alert`), and `switchyard/portal/app.py` line 276 (`threshold = reg.settings.transient_breaker.streak_alert`) — the operator-configurable value drives the threshold.

If an operator sets `streak_alert: 5`, the chip appears at streak 3 while the corresponding `alerting.append(...)` message in the plans table waits until streak 5, so the same operator-facing surface disagrees with itself: a row can show a `failing · 4x` warn chip but no plan-row alert, exactly the in-between state the docstring describes as "the operator's cue".

Suggested fix: thread the threshold into the capacity context (e.g. add `streak_alert` next to `transient_streak` on the rows in `picker.capacity`, or hoist it onto the `capacity` dict in `collect_capacity`), then make the template compare against that value. A test that temporarily raises `streak_alert` and checks the chip is absent at streak 3 would lock this in — the current `test_capacity_row_shows_failing_chip_when_streak_meets_alert` only exercises the default 3, so it can't tell the two apart.

switchyard/picker.py
```diff
@@ -122,9 +129,15 @@ async def pick(self, lane: str, session: str | None,
 
         # 1. Affinity. A session that already has a provider stays on it as
         #    long as that provider is still in the lane and has a free slot.
+        #    On a re-pick the held lease may be the member that just failed;
+        #    skip the pin in that case (do NOT drop_lease — a transient failure
+        #    must not strand the session on a peer) and let ordered fill land
+        #    on whoever has room. The successful re-pick's set_lease moves the
+        #    session to its new plan, and pinned semantics are otherwise
+        #    untouched.
```
Fledgewing: **[should-fix] The "do NOT drop_lease" comment contradicts the code**

The comment block at lines 130-137 asserts:

> On a re-pick the held lease may be the member that just failed; skip the pin in that case **(do NOT drop_lease — a transient failure must not strand the session on a peer)** and let ordered fill land on whoever has room.

But lines 190-191 do exactly that:

```python
elif held:
    await self.slots.drop_lease(session)
```

The elif runs whenever `held and (held not in by_ref or held in exclude)`, i.e. both "the held plan is no longer in the lane" and "the held plan is the one we just excluded from the re-pick". The end state the test asserts (`held_after == picked.ref`, see `test_repick_moves_the_session_lease_and_returns_not_sticky`) still holds because the successful re-pick's `set_lease` overwrites the dropped value, but the comment makes a behavioral claim the code does not keep.

Either fix the comment to match the code (and acknowledge that the drop_lease is functionally redundant — the subsequent set_lease would re-establish the lease anyway), or remove the elif's drop_lease so the comment's claim becomes true. The current state is a documented invariant the next maintainer will rely on that the implementation does not deliver.

switchyard/gen_litellm.py
```diff
@@ -128,15 +128,27 @@ def build(plans_path: str) -> dict:
...
+            "num_retries": 1,
         },
         "router_settings": {
             "enable_pre_call_checks": True,
             "routing_strategy": "simple-shuffle",
             "fallbacks": fallbacks,
             "context_window_fallbacks": context_fallbacks,
-            "allowed_fails": 1,
-            "cooldown_time": 60,   # short: the real cooldowns live in Redis
+            # Load-bearing with allowed_fails=1: litellm would otherwise cool
```
Fledgewing: **[nit] The "Load-bearing with allowed_fails=1" comment is stale**

The comment block at lines 145-150 reads:

> Load-bearing with allowed_fails=1: litellm would otherwise cool the single-deployment group between attempts and the re-pick would never fire. cooldown_time:0 does NOT work — litellm's router takes the `or` default, not our zero.

But `allowed_fails` is no longer set in this file (the PR removed it). With `disable_cooldowns=True` alone, the cooldown ADD logic in `litellm/router_utils/cooldown_handlers.py` short-circuits regardless of `allowed_fails`, so the "load-bearing" relationship the comment describes is gone — the new behavior is "disable_cooldowns replaces the previous allowed_fails + cooldown_time pair". The follow-on sentence about `cooldown_time:0` is also obsolete, because `cooldown_time` is no longer emitted at all.

A reader trying to verify "is this still right?" can't, because the comment names settings that are not in the file. Suggest rewriting the block as: "`disable_cooldowns: True` is the only setting that keeps the router from cooling the single-deployment group between attempts; SwitchYard's own cooldowns (via Redis) are the only state that outlives the process and the only state the portal can see." `tests/test_routing.py:551-552` repeats the same wording and could be edited in the same pass.

switchyard/slots.py
```diff
@@ -190,3 +196,33 @@ async def touch_lease(self, session: str, ttl: int) -> None:
...
     async def drop_lease(self, session: str) -> None:
         await self.redis.delete(K_LEASE.format(session=session))
+
+    # -- transient-failure streak -----------------------------------------
+    # The escalated-cooldown ladder and the board's failing chip both need to
+    # know how many TRANSIENT failures in a row a plan has produced. We track
+    # that here so the hook does not have to invent a counter, and so a plan
+    # that was merely quiet for two hours does not look like it has been
+    # failing this whole time.
+    async def note_transient_failure(self, plan: str) -> int:
```
Fledgewing, line 258: **[nit] `note_transient_failure` is not atomic — INCR + EXPIRE can leak a key without TTL**

```python
value = await self.redis.incr(key)
await self.redis.expire(key, 7200)
return int(value)
```

The comment above the function is right about the intent ("a quiet gap of TTL seconds drops the counter entirely"), but `INCR` does not set a TTL on a new key, and `INCR` does not refresh the TTL on an existing one — only `EXPIRE` does. If the gateway worker is killed between the `INCR` and the `EXPIRE` of the very first failure for a plan, the counter exists with no TTL, so the "sliding window drops the counter" property is broken for that key. Subsequent failures heal it (because `EXPIRE` is called every time), and a TTL-less key is no worse than a TTL'd one until the process restarts, so the practical impact is small — but it is also pointless to leave on the table, since `SET NX EX 7200 0` then `INCR`, or a single `EVAL` of `INCR` + `EXPIRE`, removes the gap.

The same applies, more importantly, to the INCR-then-`cool_down` pair in `_apply_verdict`: two concurrent TRANSIENT failures on the same plan can land in either order, and the last `cool_down(plan, ...)` write is what the picker observes even though it was computed against a smaller `streak` than the final `INCR` value. A single `EVAL` that increments the streak AND writes `K_COOL` in one round-trip keeps the cooldown and the streak in lock-step.

Fledgewing:
Reviewed 4 findings on PR #42 at 448bf6d: 0 blocker, 2 should-fix, 2 nit — see inline comments at 448bf6d.

PR #50, opened 2026-09-22
picker: expire stale reported percentages at reset_at / period rollover (#45)

### LINKED ISSUE #45: Plans with no utilization probe stay spent forever after a quota rollover (Claude Code / Codex)
## Summary

A plan that gets its utilization from the vendor's own client rather than a probe (Claude Max via the Claude Code CLI, Codex) can never come back from hitting 100%. When its window rolls over and capacity is available again, the gateway still treats it as spent — and because the only thing that could refresh the reading is real traffic, which the gateway refuses to send, it stays dead until someone clears Redis by hand.

## Why it sticks

1. For these plans the only source of a usage percentage is the vendor's own client reporting during a real request (`sidecars/cli_bridge/server.py`, `usage_report()` — deliberately never a direct call to the vendor's usage API). No probe poller writes fresh readings for them.
2. When the plan runs out, `reported_pct_used: 100.0` is written via `note_reported_percent()` (switchyard/usage.py:265) — along with `reported_at` and, when the client gives one, `reset_at`.
3. `Picker.is_spent()` (switchyard/picker.py:126) returns `pct >= 100.0` with **no staleness or rollover check**: it never compares `reported_at`/`reset_at` against "now". `_cap()` then returns `0, "quota spent"` (switchyard/picker.py:121) and the plan is excluded from every pick.
4. After the window rolls over the stale 100% is still in Redis. The plan gets zero traffic, so the vendor client never reports a fresh (lower) percentage, so `is_spent()` stays True. Permanently.

Plans with a real probe self-heal, because the poller (switchyard/probes.py) keeps writing fresh readings past the rollover. The classify-driven `quota_exhausted` cooldown also recovers on its own — the stuck path is specifically the reported-percentage one.

## Suggested fix

- In `is_spent()` (and `window_headroom()`), treat `reported_pct_used` as expired once `reset_at` has passed — or once `reported_at` is older than the window's period — so a stale reading means "unknown", not "spent", and the plan re-enters rotation.
- Optionally, a cheap canary after rollover would re-learn the real number on the first request instead of waiting for the vendor client to volunteer one.

Fledgewing:
Fix #45. Plans that get their utilization from the vendor's own client during a real request (Claude Max via the Claude Code CLI, Codex) used to stay spent forever after their window rolled over: `Picker.is_spent()` trusted `reported_pct_used` / `reported_remaining` indefinitely, `_cap()` then returned `0, "quota spent"`, the plan got zero traffic, so the vendor client never wrote a fresh reading — a permanent over-exclusion bug.

The fix is read-side only. A new helper `switchyard.usage.reported_is_current(facts, period, now=None)` decides whether a `sy:qwin:...` reading is still meaningful: a reading is current iff its `reset_at` (when present as a float) is still in the future AND its `reported_at` is inside the current period bucket via `period_bounds(period)`. Anything undatable — missing `reported_at`, `reset_at == ""`, or a past reset — is stale. Three reads now gate on it: `Picker.is_spent()` (stale → not spent → re-enters rotation), `usage.window_headroom()` (stale → falls back to the ledger / observed-allowance basis so the board shows the estimate), and `portal._recompute_perishable_for_plan` (stale → member scores `None` and sorts under "unknown"). Writers, Redis keys, the `quota_exhausted` cooldown, the pacer, and both sidecars are untouched.

Tests: five new cases across `tests/test_usage.py`, `tests/test_routing.py`, and `tests/test_portal.py` — expiry on past `reset_at`, expiry at period rollover, headroom falling back to ledger basis, a routing assertion that a spent plan re-enters rotation after the window rolls over, and the perishable writer treating stale target + stale gate as unknown.

One small test-data refresh: `tests/test_probes.py::test_minimax_remains_percent_payload_is_read_as_percentages` had three `end_time` / `weekly_end_time` fields hardcoded to Sep 21 2026 (now ~27h in the past). The new gate correctly identifies these as stale and exposes the pre-existing fixture brittleness; bumping those three ms values to Sep 22 2026 (with `weekly_end_time` still satisfying the test's `1.79e9 < reset_at < 1.8e9` bound) makes the suite green again. No assertions changed.

Note for the reviewer: `switchyard/*.py` is baked into the gateway/portal/sidecar images (not mounted), so `scripts/apply.sh --build` from the main checkout is the user's manual step after merge — not part of this PR.


tests/test_probes.py
```diff
@@ -304,12 +304,12 @@ def test_minimax_remains_percent_payload_is_read_as_percentages():
...
+         "start_time": 1789948800000, "end_time": 1790086380000,
          "current_interval_total_count": -1, "current_interval_used_count": -1,
          "current_interval_used_percent": "37.5%",
-         "weekly_start_time": 1789948800000, "weekly_end_time": 1790553600000,
+         "weekly_start_time": 1789948800000, "weekly_end_time": 1790121600000,
          "current_weekly_used_count": -1, "current_weekly_used_percent": "12%"},
-        {"model_name": "video", "end_time": 1790035200000,
+        {"model_name": "video", "end_time": 1790086380000,
          "current_interval_used_percent": "0%"},
     ]}
```
Fledgewing, line 318: **[should-fix] Fixture refresh is time-bombed — please make the timestamp relative**

`end_time: 1790086380000` and `weekly_end_time: 1790121600000` are static epoch ms hardcoded to roughly Sep 22 2026 14:13 UTC and Sep 23 2026 00:00 UTC. Right now (Tue Sep 22 2026 ~08:30 UTC) they are ~5.7h and ~15.5h ahead of `time.time()` respectively. As soon as wall clock passes them, the new gate marks the reading stale again — `reported_is_current({reset_at: past}, "week")` returns False — and `window_headroom` falls back to the ledger basis (0% pct_used), failing the `got["weekly"].used_percent == 12.0` and `windows["weekly"]["pct_used"] == 12.0` assertions. So this fixture is back to red within ~24h.

The PR description says "no assertions changed," which is true, but the choice of Sep 22 is what makes it brittle. Three options, any of which would survive without manual bumps:

  1. Pick a far-future timestamp that still fits the assertion `1.79e9 < reset_at < 1.8e9`. The upper limit `1.8e9` is Mar 2027, so any value up to that is fine and always future-dated from today.
  2. Compute the timestamps from `time.time()` inside `go()`, e.g. `end_time = int(time.time() * 1000) + 6 * 3600 * 1000` — always N hours ahead.
  3. Pin the test's clock via the `_patch_now` helper already used in `tests/test_usage.py` (you'd need to refactor `headroom`/the probe writer to read from an injectable clock, or just patch `time.time`).

Option 1 is the smallest change and matches the test's stated invariant (the actual value is irrelevant, only that the assertion range is satisfied).

switchyard/portal/app.py
```diff
@@ -184,24 +185,42 @@ async def _recompute_perishable_for_plan(reg, ledger, plan) -> None:
...
+    gate_current = (reported_is_current(gate_facts, gate_quota.period)
+                    if gate_quota else False)
+    gate_pct = (_as_float(gate_facts.get("reported_pct_used"))
+                if gate_current else None)
+    # Stale gate stops filtering members out of perishable lanes: the picker
+    # only trusts `gate5h` as a gate, and an old `>90%` row is the kind of
+    # false positive that quietly suppresses a plan until its next poll.
     this_plan_gate5h = gate_pct is not None and gate_pct > 90
     scoped = await ledger.window_facts(plan.key, "weekly_scoped")
+    scoped_current = reported_is_current(scoped, "week") if scoped else False
```
Fledgewing: **[nit] `if scoped else False` guard is redundant**

```python
scoped = await ledger.window_facts(plan.key, "weekly_scoped")
scoped_current = reported_is_current(scoped, "week") if scoped else False
```

`reported_is_current({}, "week")` already returns False on its own: there is no `reported_at` (rule (b)'s `not isinstance(reported_at, float)` short-circuits to False), so the explicit `if scoped else False` branch never returns anything different. Drop the guard for symmetry with the target/gate paths above, which both trust `reported_is_current` to decide without caller-side truthiness:

```python
scoped_current = reported_is_current(scoped, "week")
```

Fledgewing:
Reviewed 2 findings: 0 blocker(s), 1 should-fix, 1 nit — see inline comments at 5a0ca20.

Quick summary of what I read and what I think:

* The gate logic is correct. `reported_is_current` cleanly separates rule (a) — `reset_at` past → stale, including the `note_exhaustion` `""` case via the `not isinstance(..., float)` short-circuit — from rule (b) — `reported_at` outside the current period bucket → stale. The three read sites (`Picker.is_spent`, `window_headroom`'s two early-return branches, and `portal._recompute_perishable_for_plan`'s target/gate/scoped) all delegate to it consistently, and none of them regress the happy path: `test_spent_plan_reenters_after_window_rollover`'s control case (future `reset_at` + 100%) still skips. Deferred canary from the issue is fine — over-admission self-heals via the vendor's next 429 → cooldown, so the canary is a speed-up, not a correctness fix.
* Test coverage is good: five new cases across `test_usage.py`, `test_routing.py`, `test_portal.py` cover rule (a), rule (b), `window_headroom` falling back to ledger, picker re-admitting after rollover, and the perishable writer dropping both target and gate. All tests are inserted before the `__name__ == "__main__"` runner as the repo's CLAUDE.md requires.
* The fixture refresh in `tests/test_probes.py` (should-fix, posted) is the brittle part: bumping `end_time`/`weekly_end_time` to ~Sep 22 2026 puts the test back to red within ~24h. Suggest picking a far-future value still inside `1.79e9 < reset_at < 1.8e9` rather than the immediate-future pattern that has now been done twice.
* One nit: `switchyard/portal/app.py` line 221 has a redundant `if scoped else False` guard around `reported_is_current` (function already returns False for `{}`), worth simplifying to match the surrounding target/gate patterns.

Issue #52, opened 2026-09-22
cli_bridge and mcp_bridge have no /v1/messages route, so Claude Code 404s on every CLI seat


subconsciousbias:
Same reporter as #44. Running real Claude Code tabs against the gateway.

## What happens

LiteLLM forwards an Anthropic-shaped body straight through to the deployment. `sidecars/token_proxy/server.py` has `@app.post("/v1/messages")`, but `cli_bridge` and `mcp_bridge` do not, so an Anthropic-protocol client pointed at a CLI seat gets a 404 from the sidecar. Claude Code itself is that client, which makes the CLI seats unreachable from the tool the whole thing exists to serve.

There is a second 404 in front of it: an `openai/`-prefixed deployment is reached through the Responses API, so LiteLLM POSTs `/v1/responses`. Declaring the endpoints the sidecar really serves fixes that, on both the per-model entries and the lane entry in `gen_litellm.build`:

```python
if plan.auth in ("cli_sidecar", "oauth_sidecar"):
    info["supported_endpoints"] = ["/v1/chat/completions", "/v1/messages"]
```

## What I did here

Added the route to both bridges, about 160 lines each:

- `_anth_to_openai` / `_openai_to_anth` translate the body and the reply, `_anth_sse` frames the stream
- the route calls the same `_handle_chat` the chat route calls, so slots, cooldowns, stickiness and every limit behave identically. No second path to keep in sync.
- `claude -p` yields a whole answer rather than a token stream, so a streamed request goes out as one content block plus the closing events. The right shape, just not incremental. Clients accept it.

Tabs have been running tool loops through this for a day.

One caveat if you take it: this route does not fire LiteLLM's success logging event, which leaks the slot. Filed separately as #51, since that one hits `token_proxy` today too and is worth fixing first.

Happy to open a PR for either or both.

PR #54, opened 2026-09-22
hooks: release the slot and book the tokens on routes LiteLLM does not log

### LINKED ISSUE #51: /v1/messages fires no success logging event: the slot is never released and the tokens are never booked
Same reporter as #44. I found this while adding the Anthropic route to the bridges, but this half is upstream already: `token_proxy` serves `/v1/messages` today, so any traffic on that route is affected.

## What happens

`async_log_success_event` is where the slot release and the ledger write live, and it fires for `completion` / `acompletion` / `text_completion` / `atext_completion`. It does not fire for `/v1/messages` at all. So a request on that route:

- holds its slot until the staleness sweep picks it up
- spends quota the ledger never sees, so the portal under-reports that plan

Measured here: a Claude Code tool loop filled a 2-slot seat and then refused its own follow-up call, with no real work in flight on that seat.

Streaming is worse. A streamed reply never reaches `async_post_call_success_hook` either, so on that route there is no place left at all to finish the request.

## What I did here

- stash `call_type` in the pick context at claim time, since the finish has to know which path it is on
- add `_finish_unlogged(ctx, response)`: if `call_type` is not one of the four logged types, stop the heartbeat, release the slot, read usage, write the ledger, note throughput. Anthropic names its counters `input_tokens` / `output_tokens` and a translated reply may still carry the OpenAI pair, so it reads either.
- call it from `async_post_call_success_hook`, and from a `finally` in `async_post_call_streaming_iterator_hook`, so a client that hangs up mid-stream does not keep the seat

Allow-list rather than deny-list on purpose: a future route that does not log then fails safe.

About 60 lines in `hooks.py`, and it does not depend on anything else I have. Happy to open a PR if you want it.

subconsciousbias:
Fixes #51.

### The bug

`async_log_success_event` only fires for `completion`, `acompletion`, `text_completion` and `atext_completion`. Three things live in that event: the slot release, the ledger write, and the pacer throughput sample. So a request that arrives on any other route gets none of them.

`/v1/messages` is that route today, and `sidecars/token_proxy` already serves it, so this is reachable on `main` right now without any of the bridge work in #52.

What it looks like from the outside: the seat stays claimed until the staleness sweep, and the tokens it spent never reach the ledger, so the portal under-reports that plan. Measured here against a Claude Code tool loop - it filled a 2-slot seat and then refused its own follow-up call, with no real work in flight on that seat.

### The fix

Stash the call type in the routing context at claim time, then finish by hand in the post-call hooks when the logging event is not going to fire.

Three notes on the shape of it:

- **Allow-list, not deny-list.** `_LOGGED_CALL_TYPES` names the call types that *are* logged, so a future route that does not log fails safe - it gets finished by hand - instead of silently leaking a slot.
- **The streamed path finishes in a `finally`.** A streamed reply never reaches `async_post_call_success_hook`, and a client that hangs up mid-stream must not keep the seat.
- **Usage is read under either name.** Anthropic calls its counters `input_tokens` / `output_tokens`; a translated reply may still carry the OpenAI pair. It reads both, and a missing usage block books zero rather than raising.

`_split_reasoning` is pulled out into its own method so the iterator hook can wrap both paths in one `try`. No behaviour change there.

### Blast radius

Nothing changes for a chat-completions request - it returns at the first line of `_finish_unlogged`. One file, +58/-3.

### Testing

Verified against a live Claude Code session through the relay: the slot is released when the turn ends rather than at the sweep, and the plan's token counters move on `/v1/messages` traffic where they previously stayed flat.


subconsciousbias:
Pushed a second commit: the first cut released the slot on the streamed path but finished with no response object, so every streamed turn booked zero tokens.

Caught it by looking at the ledger rather than the code - the slot was coming back, but `sy:usage:<plan>:h:*` for the plan serving a live Claude Code session had not moved in six hours of continuous traffic, while a non-streaming plan on the same gateway was booking normally every hour.

Both bridges put the usage block on the final chunk beside `finish_reason`, so the iterator hook now keeps whatever the stream last carried and finishes with that. +22/-1 on the same file.

subconsciousbias:
Pushed a third commit. The slot half of this PR was right, but what it booked was not - twice over. Both were found by reading the ledger this gateway actually wrote, not the code that writes it.

### 1. The cache counters are the prompt

A cached Anthropic turn reports almost nothing in `input_tokens`. Measured here on one real Claude Code turn through the bridge:

```
"usage": {"input_tokens": 6,
          "cache_creation_input_tokens": 6932,
          "cache_read_input_tokens": 41302,
          "output_tokens": 201}
```

48,240 prompt tokens. `_finish_unlogged` booked **6**. An hour of that traffic in `sy:usage:claude-max:h`:

| hour (UTC) | requests | prompt_tokens | completion_tokens |
|---|---|---|---|
| 09-21T12 | 170 | 472 | 7476 |
| 09-22T08 | 159 | 0 | 0 |
| 09-22T09 | 125 | 0 | 0 |

The zero hours are fault 2 below. The 472 is this one: 170 requests, 472 prompt tokens between them, and that is the number the capacity planner spends the subscription against.

`prompt_tokens` stays authoritative wherever it exists - cli_bridge already folds both cache counters into it before it answers, and `prompt_tokens_details.cached_tokens` is a subset of it - so the counters are only added when `prompt_tokens` is absent. Nothing double-counts, and the metered providers are untouched.

### 2. `_note_usage` kept the wrong half

My own commit from earlier today. It keeps the last usage block the stream carried, which is right for the OpenAI framing - `sse_from_completion` emits it once, complete, at the end - and wrong for Anthropic, which splits it:

* `message_start` -> `input_tokens` + both cache counters
* `message_delta` -> `output_tokens`, near the end

Last-block-wins keeps the second one, so a streamed turn booked the entire prompt as zero. That is the `0 / 0` in the table above, against 159 requests in an hour. It is now merged field by field.

### Tests

Both are invariants rather than cases, and I ran them against the previous commit of this branch to be sure they bite:

* `test_a_cached_prompt_is_booked_whole` - the 48,240 figure above, plus the both-shapes case that would double-count.
* `test_a_split_stream_keeps_both_halves_of_its_usage` - message_start then message_delta, asserting prompt 48,240 and completion 201.

Worth saying plainly: without the slot fix in this PR the seat leaks, and with it the ledger was wrong in a way that looked like the plan was free. The second is the quieter of the two.

zeroh-labs:
Read this diff and #74's, which merged an hour ago. Three things that are decidable from the two diffs alone, ordered by how hard they are to see later.

**`_finish_unlogged` now has two call sites for one request and no idempotency guard.** It is awaited from `async_post_call_success_hook` and from the streaming iterator's `finally`. Each call does `_stop_heartbeat`, `picker.release`, `ledger.record` and `note_throughput`, and nothing in `ctx` records that it already ran. So the correctness of the booking rests on "at most one of those two paths fires per request", which is a property of LiteLLM's hook ordering rather than of this file. I can't settle from the diff whether `async_post_call_success_hook` fires for a streamed reply — but the function doesn't need that to be true. A single flag makes it stop mattering:

```python
if ctx.get("finished"):
    return
ctx["finished"] = True
```

Worth having because the failure is silent in the direction that matters: a second `ledger.record` double-books the turn, and a second `release` returns a seat that may already have been reclaimed.

**The two functions in this PR disagree about which name wins for completion tokens.** `async_log_success_event` reads `get("completion_tokens") or get("output_tokens")`; `_finish_unlogged` reads `get("output_tokens") or get("completion_tokens")`. Opposite precedence, same payload family. It is unobservable while only one name is present — but `_USAGE_FIELDS` lists both and `_note_usage` stores every truthy field it sees, so the merged `seen` dict from a translated stream can carry both at once. Then the same turn books one number if it finished through the logging event and the other if it finished by hand, and the divergence lives in whichever counter the two producers disagree about.

**`_prompt_tokens`' gate is presence, not wholeness — and main now has a second copy of it.** #74 merged the same rule into `_prompt_completion_tokens` (`if src.get("prompt_tokens")` → skip the cache fold), so once this rebases the file holds two independent implementations of one discriminator, reading usage from different sources. Both share the same blind spot: presence of `prompt_tokens` is a fact about the producer, wholeness is a fact about the value, and the tests on both sides only cover payloads where those coincide.

There is an arithmetic tell in the payload that doesn't need the producer assumption. A folded total is at least the sum of its cached components:

- `prompt_tokens >= cache_read + cache_creation` → consistent with folded, honour verbatim.
- `prompt_tokens < cache_read + cache_creation` → cannot be the folded total, so fold.
- equality is the one genuinely ambiguous case (it is what `input_tokens == 0` looks like either way).

On your own fixture: 48240 ≥ 48234 passes through untouched, while `{prompt_tokens: 6, cache_read: 41302, cache_creation: 6932}` books 48240 instead of 6. That second shape is the same three-orders-of-magnitude undercount your `test_a_cached_prompt_is_booked_whole` docstring is about, arriving through the other door — and after #74 the residual risk on both paths points that way rather than at the double count, which is the harder direction to notice because it looks like headroom.

I'm going on `cli_bridge`'s docstring for what `to_openai` emits, not the bridge source, so I'm not asserting a live instance of that shape exists in the tree today.

I can write the guard plus a fixture for the third shape against either function, and the precedence fix is one line once you pick which name wins.

subconsciousbias:
closing, main now releases the slot and books /v1/messages itself (#62, #67)

Issue #47, opened 2026-09-22
Capacity 429s surface to callers: litellm never invokes SwitchYard's re-pick hook, so the router's one retry re-hits the same deployment


Fledgewing:
## Symptom

OpenCode surfaced this twice today on `judge` traffic:

```
litellm.RateLimitError: RateLimitError: OpenAIException - Error code: 429 -
{'detail': 'sidecar at capacity (4)'}. Received Model Group=sy.opencode-go.glm-5.3-flash
Available Model Group Fallbacks=None
```

Expected behaviour: the sidecar says "full", the router's single retry re-enters SwitchYard's picker, which spills to `opencode-go2/glm-5.3-flash` (or `glm/glm-5.3-flash`). Actual behaviour: the retry re-sent the request to **the same deployment**, got the same 429, and litellm gave up and handed the raw error to the caller.

## Root cause

The lane-spill-on-retry design (gen_litellm.py: general `fallbacks` are deliberately `[]`, `num_retries: 1`, and `SwitchyardHandler.async_pre_routing_hook` is supposed to swap in a different lane member on the retry) depends on litellm invoking `CustomLogger.async_pre_routing_hook`. **The litellm build in the gateway image never does.**

Evidence from the running gateway (`ghcr.io/berriai/litellm:main-stable`, pulled today):

- `grep -rn pre_routing_hook` over `site-packages/litellm` shows the only dispatch sites are litellm's own internal router-strategy plugins (`_select_pre_routing_strategy` → adaptive/quality/auto/complexity routers, router.py:13487). `Router.async_pre_routing_hook` never iterates registered callbacks; `CustomLogger.async_pre_routing_hook` (custom_logger.py:250) is a dead override target in this build.
- `docker logs switchyard-gateway-1`: the hook's log line `lane=... re-picked after transient ...` has **never** fired, while `lane=judge -> opencode-go/...` pick lines and failure-logging errors appear constantly.
- With `fallbacks: []` (by design — see the comment in `switchyard/gen_litellm.py`) and a single-deployment group, the retry has nothing else to pick, so litellm reports exactly what we see: `Model Group=sy.opencode-go.glm-5.3-flash`, `Available Model Group Fallbacks=None`.

[...]

PR #55, opened 2026-09-22
Fix _apply_verdict UnboundLocalError: bind cooldown from verdict (#48)

### LINKED ISSUE #48: _apply_verdict raises UnboundLocalError on every coollable failure — no cooldown or breaker streak has been applied since #42
## Symptom

```
...
associated with a value
```

## Root cause

```python
        did_cool_atomic = False
            if breaker.enabled:
                    plan.key,
                    cap=breaker.max_seconds,
                    reason=verdict.outcome.value,
                )
```

and the fallback path:

```python
```

## Why the tests missed it

## Fix

   ```python
   ```

Fledgewing:
## What broke

The #42 refactor (`Retry once on transient upstream failures; escalate repeated cooldowns`) renamed the cooldown use sites in `_apply_verdict` but dropped the `cooldown = verdict.cooldown_seconds` binding. After that landed, **every** coollable verdict (RATE_LIMITED, TRANSIENT, QUOTA_EXHAUSTED, PLAN_DEAD, AUTH) raised `UnboundLocalError: cannot access local variable 'cooldown'` the first time it tried to cool a plan.

The damage was quiet because the ledger counts (`note_concurrency_rejection`, `note_reported`) still landed *before* the cooling section crashed — so the visible symptom was missing cooldowns, not stack traces. In particular:

- The **TRANSIENT** path never called `bump_and_cool`, so the transient-failure streak was never bumped and the escalated ladder never fired.
- **RATE_LIMITED** / **QUOTA_EXHAUSTED** / **PLAN_DEAD** / **AUTH** paths never wrote their `sy:cool:` key.
- **QUOTA_EXHAUSTED** / **PLAN_DEAD** / **AUTH** never called `drop_lease`, so a session stayed pinned to dead capacity instead of re-leasing onto a peer.

This is the cooldown gap behind the lane-spill incident in #47: traffic kept re-feeding dead or rate-limited plans because nothing ever told the picker to back off.

## The one-line fix

`switchyard/hooks.py` — bind `cooldown = verdict.cooldown_seconds` immediately before `did_cool_atomic = False` (line 661). This mirrors pre-#42 behaviour and feeds both branches of the cooling section:

- The `TRANSIENT` branch reads it as `base=cooldown` for `bump_and_cool`, then recomputes via `escalated_cooldown(cooldown, streak, cap=...)`. The atomic INCR/EXPIRE Lua already used `cooldown` correctly — the binding is what gave it a value to start from.
- The `cool_down` branch (`if verdict.should_cool and not did_cool_atomic`) reads it directly. The `drop_lease` branch is downstream of the binding and was never missing the variable — it just never ran because the cooling section crashed first.

Why that exact binding point: pre-#42 the binding lived at the top of the "Apply the cooldown" comment block, and the `TRANSIENT` ladder was preserved via `escalated_cooldown`/`bump_and_cool`. The fix restores that position verbatim.

## Tests

`tests/test_verdict.py` — new file, five FakeRedis end-to-end tests driving `_apply_verdict` directly (no Picker, no provider, no real Redis):

1. `test_transient_with_breaker_writes_cooldown_key_and_respects_cap` — eight consecutive TRANSIENT failures; the streak grows 1..8 and the cooldown lands at the breaker cap (1800s) once `60 × 2^(n-1)` exceeds it. **Regression proof**: with the `cooldown = ...` line stashed, the first call raises the original `UnboundLocalError`.
2. `test_transient_without_breaker_uses_cool_down_with_verdict_cooldown` — breaker disabled, `bump_and_cool` is skipped, the verdict's own cooldown lands via `cool_down` and the streak stays at 0.
3. `test_rate_limited_writes_cooldown_key_and_raises_no_exception` — RATE_LIMITED writes the `sy:cool:` key at the verdict's value; the test explicitly asserts no exception escapes (the Litellm callback would otherwise swallow a bug here).
4. `test_quota_exhausted_writes_cooldown_and_drops_lease` — QUOTA_EXHAUSTED both cools the plan **and** drops the session lease, so the next pick re-leases onto a peer rather than re-feeding dead capacity.
5. `test_double_count_guard_skips_second_call_with_same_request_id` — same `ctx.request_id` short-circuits the second `_apply_verdict`; a fresh `request_id` is allowed through. Without the per-attempt marker, the streak would double and the cooldown would jump straight to the cap on a single 5xx.

## Verification

Offline suite, all green from this worktree:

```
$ python3 tests/test_verdict.py
... 5 verdict tests passed

$ python3 tests/test_routing.py tests/test_classify.py 2>/dev/null
... all routing tests passed

$ for t in tests/test_*.py; do python3 "$t" >/dev/null || echo "FAIL $t"; done
(no FAILs)

$ python3 tests/render_preview.py
rendered index.html + fragments (87900 bytes) -> /tmp/switchyard-preview.html
  10 plans, 19 models, 5 lanes, 18 quota windows, 4 probes
```

## Deploy note

`switchyard/` is `COPY`ed into the gateway / portal / sidecar images — only `./config` is mounted from the host. This commit is **inert until the main checkout rebuilds and redeploys**:

```bash
# from the main checkout, not from this worktree
docker compose build gateway portal claude-max-sidecar xai-token-proxy
docker compose up -d
```

Closes #48.


tests/test_verdict.py
```diff
@@ -0,0 +1,298 @@
...
+# someone's real config would otherwise silently become the fixture.
+os.environ["SWITCHYARD_PLANS"] = plans_path()
+
+from dataclasses import replace  # noqa: E402
+
+from switchyard import models  # noqa: E402
+from switchyard.classify import Outcome, Verdict  # noqa: E402
+from switchyard.hooks import SwitchyardHandler  # noqa: E402
+from switchyard.policy import CapacityPolicy  # noqa: E402
+from switchyard.slots import K_COOL, K_LEASE  # noqa: E402
```
Fledgewing: **[nit] Unused imports K_COOL / K_LEASE**

Neither name is referenced anywhere in the test file — `K_COOL` only shows up in a comment (line 234) and `K_LEASE` doesn't appear at all. The cooldown and lease shapes are already covered transitively by `slots.cooldown_state(...)` and `slots.get_lease(...)`, so the explicit constants add noise without sharpening any assertion. Other tests in the suite (e.g. `test_routing.py`) match this convention and don't import them either. Drop both names from the import line.

Fledgewing:
Reviewed 1 finding: 0 blocker(s), 0 should-fix, 1 nit — see inline comment at 3ab517d.

PR #56, opened 2026-09-22
hooks: bind cooldown before the verdict branches read it

### LINKED ISSUE #48: _apply_verdict raises UnboundLocalError on every coollable failure — no cooldown or breaker streak has been applied since #42

subconsciousbias:
Fixes #48.

`cooldown` is read twice in `_apply_verdict` - as `base=` for the TRANSIENT breaker and again in the generic `cool_down` call - and nothing ever assigns it, so every coollable verdict raises `UnboundLocalError` before either one lands.

Binding it from the verdict up front, above the outcome branches, since every branch below assumes it exists.

### Seen on a second stack

Independent of the incident in the issue - a different deployment, different plans, same signature:

```
File "/switchyard/switchyard/hooks.py", line 729, in _apply_verdict
    base=cooldown,
UnboundLocalError: cannot access local variable 'cooldown' where it is not associated with a value
```

`sy:cool:*` and `sy:tfail*` are both 0 keys there too, over the whole life of the gateway, so the same conclusion holds: no cooldown has ever been written and the breaker streak has never incremented. A seat that starts failing keeps getting picked at full weight.

### On the test gap

Agreed with the suggestion in the issue, and it is worth stating as the invariant rather than as a case: after any coollable verdict, `sy:cool:<plan>` exists. That assertion fails today on every outcome class, so one test over the fake Redis covers all of them.


subconsciousbias:
Added the test you asked for, as the invariant: after any coollable verdict, `sy:cool:<plan>` exists.

Checked over the fake Redis for TRANSIENT, RATE_LIMITED, QUOTA_EXHAUSTED, CONCURRENCY and PLAN_DEAD, each on its own Redis so one class's key cannot stand in for another's, plus BAD_REQUEST which must **not** cool - our own bad prompt is not the plan's fault. Reverting the one-line bind fails it with the exact `UnboundLocalError` from the issue, on the first class it tries.

Stating it as the invariant rather than one case was deliberate: pinned to a single outcome it would read as one gap, when what was actually true is that no outcome class had ever cooled a plan at all.

subconsciousbias:
closing, #55 fixed this on main

PR #60, opened 2026-09-22
Fix #47: caller-retry spill contract, litellm v1.101.0 pin + startup self-check, CONCURRENCY classification, gate headroom

### LINKED ISSUE #47: Capacity 429s surface to callers: litellm never invokes SwitchYard's re-pick hook, so the router's one retry re-hits the same deployment

### LINKED ISSUE #48: _apply_verdict raises UnboundLocalError on every coollable failure — no cooldown or breaker streak has been applied since #42

Fledgewing:
## Problem

The gateway was surfacing raw sidecar capacity 429s to callers because the lane-spill design relied on `litellm` dispatching `CustomLogger.async_pre_routing_hook` to re-pick a different member of the lane on a transient failure. The gateway image's pinned litellm never dispatches that hook to user code: it is the internal auto-router hook, fired only inside `async_get_available_deployment` for `routing_strategy`-based strategies, and it does not iterate the `CustomLogger` callback registry. So the re-pick was a dead letter and the single `num_retries: 1` retry re-routed to the same single-member deployment that had just 5xx'd.

Three related defects compounded it:

- `_apply_verdict` crashed with `UnboundLocalError` on every coollable failure (`cooldown` referenced before assignment in `hooks.py`), so plans never actually cooled and the next request re-picked the same plan that had just failed.
- Genuine capacity refusals ("sidecar at capacity (N)") classified as `RATE_LIMITED`, starving the concurrency learner and missing the 20s sit-out.
- The gateway slot table and the sidecar's own gate shared the same cap. Any claim/release race produced a spurious 429: the gateway released a slot a moment before the sidecar noticed, and the sidecar responded "sidecar at capacity".

## The fix - four parts

1. **Restore the README-documented contract.** `num_retries: 0` in the generated litellm config, no router fallbacks, no router cooldowns. The caller's 429 retry is the spill: it re-enters the proxy through `async_pre_call_hook` and the picker places against the fresh cooldowns the failure just set, which is strictly better placement than any in-router retry could give. Deleted the dead `SwitchyardHandler.async_pre_routing_hook` and the `failed_ref` marker it read; kept the `_verdict_applied` double-count guard so a verdict cannot be booked twice on the same attempt.

2. **`cooldown` order bug.** `cooldown = verdict.cooldown_seconds` before first use in `_apply_verdict`. Cooldowns apply again; the escalating `K_TFAIL/{plan}` ladder is reachable.

3. **Pin `ghcr.io/berriai/litellm:v1.101.0@sha256:d295634e09c648dcdb72c4cc2dd226f5fb87823a73e88cbbed6f205e4deb044b`.** Tag and digest move together (documented in `README.md`); a future bump is a deliberate commit that re-runs the startup self-check. New `switchyard/selfcheck.py`, invoked by `docker/gateway-entrypoint.sh` between `gen_litellm` and `exec litellm`. Two halves:
   - **Static AST audit** of the installed litellm source: `Router.async_pre_routing_hook` contains no `CustomLogger` callback dispatch (it stays internal-only); `Router.acompletion` reaches `async_function_with_fallbacks`; `async_function_with_retries` short-circuits on `num_retries <= 0` with an early raise, not via `range(0)`; `litellm/proxy` still references `async_pre_call_hook` somewhere on disk.
   - **Loopback stub probe**: two `http.server` stubs on `127.0.0.1` (one 200, one 429 with body `{"detail": "sidecar at capacity (1)"}`); a minimal `litellm.Router` with one CustomLogger; the success case reaches `async_log_success_event` exactly once, the 429 case reaches `async_log_failure_event` exactly once, the 429 case lands exactly one HTTP call at the stub (no blind retry), and raises `litellm.RateLimitError`.

   Any failed check logs `CRITICAL` and exits 1; the container restart policy turns that into a visible outage rather than a silently broken proxy. No env-var escape hatch.

4. **CONCURRENCY classification + gate headroom.** `"sidecar at capacity"` is the exact phrase `mcp_bridge/server.py` and `cli_bridge/server.py` emit on a full gate; the regex now catches it and classifies as `CONCURRENCY` with the 20s sit-out, feeding the concurrency learner. New `gate_headroom_slots` setting (default 1) on `Settings`, mirrored into `config/plans.example.yaml`. For CLI-backed plans the gateway's effective cap becomes `min(cap, max(1, max_parallel - gate_headroom_slots))` so the slot table is the limiting factor (and cannot race the sidecar); API plans keep the full configured cap (their provider refuses on its own and the headroom slot would sit unused). The litellm per-deployment backstop `max_parallel_requests` is aligned via `cap_for(model, settings)` so a direct-name caller can no longer race the picker for the headroom slot.

## Evidence

- `python3 tests/test_routing.py tests/test_classify.py` → all routing tests passed
- Full suite (`for t in tests/test_*.py; do python3 "$t" >/dev/null || echo FAIL; done`) → no failures
- Per-file test counts vs `origin/main`: `test_routing.py` 37→39, `test_classify.py` 13→16, `test_policy.py` 22→27
- `python3 tests/render_preview.py` → green (10 plans, 19 models, 5 lanes, 18 quota windows, 4 probes)
- `python3 -m switchyard.gen_litellm config/plans.example.yaml /tmp/opencode/int-gen.yaml` → `num_retries: 0`, `fallbacks: []`, `context_window_fallbacks: []`, `disable_cooldowns: true`
- `python3 -m switchyard.selfcheck` → `selfcheck PASSED` (all four static audits + both dynamic probes)

## After merge (main checkout only)

```sh
docker compose build gateway && docker compose up -d gateway
docker logs switchyard-gateway-1 | grep -i selfcheck    # expect probe passing
docker exec switchyard-gateway-1 \
    python3 -c "import litellm; print(litellm.__version__)"   # expect 1.101.0
```

Fixes #47
Fixes #48


tests/test_routing.py
```diff
@@ -1685,6 +1692,131 @@ def test_perishable_one_adjacent_swap_reconciles_added_and_removed_refs():
...
+        # capacity() reads the same policy.effective() the picker uses, so it
+        # already has the headroom applied. Plan cap is per-plan; every model
+        # in this plan inherits that cap or narrows it, mirroring cap_for.
+        # The picker caps at min(model_cap, plan_cap) per the picker
+        # implementation; for the litellm backstop we want min(model_cap,
+        # aligned_plan_cap). The first model on each plan is enough to verify
+        # the alignment because the per-deployment backstop is emitted per
+        # model, and the cap_for for that model is what gets written.
+        for model in plan.models.values():
+            picker_plan_cap = plan.cap_for(model, reg.settings)
```
Fledgewing: **[should-fix] Backstop can exceed picker admit under learner pressure — test claim is misleading**

The test docstring (lines 1740-1749) and the comment block at lines 1769-1776 both say the reference value is "what the picker would admit", but the line below uses `plan.cap_for(model, reg.settings)` — which is not what the picker actually uses. The picker goes through `policy.effective(plan).cap` (picker.py:123), and `_apply_gate_headroom` only narrows when the prior cap is *above* the headroom ceiling.

For a CLI-backed plan where the learner has pulled the cap below the headroom ceiling (e.g. `max_parallel=4`, `headroom=1`, learned=2), `policy.effective()` returns 2 but `plan.cap_for(model, settings)` still returns 3. The litellm backstop ends up at 3, so a direct-name caller can fit 3 concurrent slots while the picker admits only 2 — exactly the spurious-429 race the PR's headroom story is meant to prevent, just on the learner axis instead of the fresh-plan axis.

The test exercises the example plans (claude-max with `max_parallel=2`), where the two always coincide, so it passes — but a future config that bumps a CLI plan past 2 connections and then sees the learner pull down would silently diverge.

Suggested fix: build a `CapacityPolicy` here (the imports are already present on lines 1759-1763) and compare the backstop against `policy.effective(plan).cap` for CLI plans and `plan.cap_for(model)` for API plans, OR add a parallel `test_backstop_matches_picker_admit_when_learner_pulls_cap_below_headroom` that does exercise the divergence case.

switchyard/selfcheck.py
```diff
@@ -0,0 +1,522 @@
...
+    for root, _, files in os.walk(proxy_dir):
+        # Skip pyc caches; their source is identical to the .py alongside
+        # them and reading bytecode here would be brittle across Pythons.
+        for fname in files:
+            if not fname.endswith(".py"):
+                continue
+            path = os.path.join(root, fname)
+            try:
+                with open(path, encoding="utf-8", errors="replace") as fh:
+                    if "async_pre_call_hook" in fh.read():
```
Fledgewing: **[nit] `async_pre_call_hook` presence check is a substring scan, not a dispatch check**

The audit iterates the proxy source looking for the literal string `"async_pre_call_hook"`. That makes the gate vulnerable to silent false-negatives:

- A future version that keeps the name in a docstring, a comment, or a test fixture but stops dispatching the hook would still pass the audit.
- A future version that renames the dispatch symbol (e.g. splitting pre-call from pre-call-stream, or attaching to a typed registry) would either fail noisily (good) or pass because the old name still survives somewhere on disk (bad — and hard to notice).

A tighter check would walk the AST of each proxy source file and confirm there is at least one `ast.Call` whose function name (or attribute chain) contains `async_pre_call_hook`, or at minimum grep with a stricter pattern like `\.async_pre_call_hook\s*\(` to require the call shape rather than a bare identifier. The audit's docstring already promises the louder semantic; the implementation should match.

switchyard/selfcheck.py
```diff
@@ -0,0 +1,522 @@
...
+    litellm.logging_callback_manager.add_litellm_async_success_callback(ok_probe)
+    litellm.logging_callback_manager.add_litellm_async_failure_callback(ok_probe)
+    try:
+        ok_router = _make_router(ok_base, "selfcheck-ok")
+        ok_ok, ok_exc = await _probe_once(ok_router, "selfcheck-ok")
+        # Allow any background-task fan-out to drain before we read counts.
+        await asyncio.sleep(0.1)
+    finally:
+        ok_server.shutdown()
+        litellm.logging_callback_manager.remove_callback_from_all_lists(
```
Fledgewing, line 468: **[should-fix] Dynamic probe leaks `_ProbeImpl` callbacks into litellm's global lists**

`remove_callback_from_all_lists` defaults to `require_self=True`, which only matches bound methods where `c.__self__ == obj`. The probe registers the *instance* (`ok_probe`), but the cleanup passes `ok_probe.__class__` — so the class is the lookup key, while the stored callback's `__self__` is the instance. The cleanup silently no-ops and the `_ProbeImpl` instance stays in `litellm._async_success_callback` and `litellm._async_failure_callback` for the rest of the process.

I verified this by running `python3 -m switchyard.selfcheck` and then dumping `litellm._async_success_callback`; the probe instance is still there alongside the `Router.deployment_callback_on_success` bound methods that the in-memory Routers added. In the production entrypoint the leak is harmless — selfcheck runs in a separate process before `exec litellm`, so the new process never sees it. The leak does bite in any path that imports selfcheck in-process alongside a real litellm stack (e.g. a future test or one-shot script that wants both the audit and a live router in the same interpreter), where the probe would start intercepting every request that flows through litellm.

Suggested fix: pass the instance (`ok_probe`, `rl_probe`) rather than the class, or pass `require_self=False` to `remove_callback_from_all_lists`. The Router instances that `add_deployment` chains into the same lists also leak here — same fix, just acknowledge that those need an explicit teardown if you ever want this to be reusable.

Fledgewing:
**Two stale references that the diff doesn't touch**

These are the two cleanup items I couldn't anchor to a diff line because the files aren't in this PR's diff, but they are directly tied to the deletion the PR performs and would be a one-line follow-up to keep things honest:

1. `switchyard/picker.py:157-160` — the comment on the `exclude=` branch still claims it exists "so the router's one-shot retry sees a different candidate". With `num_retries=0` and the `SwitchyardHandler.async_pre_routing_hook` removed, no production code passes `exclude` any more (only tests do). Either drop the parameter and the dead branch, or rewrite the comment to say the exclusion path exists for the re-pick the caller's retry performs on its way back through `async_pre_call_hook`.

2. `switchyard/portal/app.py:443` — `"cap": plan.cap_for(m)` is called without `reg.settings`, so for CLI-backed plans it returns the raw `max_parallel` while the plan's `capacity` (from `policy.effective(plan)` a few lines above) returns the headroom-reduced value. The visible badge is gated by `m.narrowed`, so on the current `config/plans.example.yaml` only plans with a model-narrowed cap above the headroom ceiling would show a different number, but the conceptual gap is real and `gen_litellm.py` is now using the settings-aware form. Passing `reg.settings` here would mirror that.

Neither blocks the headline fix; both are documentation/consistency tidy-ups that line up with the contract this PR is establishing.

Fledgewing:
Reviewed 5 findings at HEAD 5a7aa53: 2 should-fix, 1 nit on selfcheck, 1 nit + 1 nit on stale references — see inline comments.

Headline contract (caller-retry spill, hook deletion, pin+selfcheck, CONCURRENCY classification, gate headroom with floor=1) is implemented as described and the offline suite is green; render_preview passes; selfcheck PASSED on the sandbox litellm 1.102.0 (image pins 1.101.0). The two should-fix items are about correctness gaps in the new checks, not regressions in the headline fix:

- backstop-vs-picker-admit gap on learner pull-down ([tests/test_routing.py:1778](#discussion_r4072097456))
- selfcheck probe callbacks leak into litellm global state ([switchyard/selfcheck.py:403](#discussion_r4072105988))

Nits:
- presence-only AST scan ([switchyard/selfcheck.py:242](#discussion_r4072100536))
- picker.py:157-160 stale router-retry comment and portal/app.py:443 `cap_for` without settings ([general comment](https://github.com/Fledgewing/SwitchYard/pull/60#issuecomment-5777321873))

No findings on the litellm pin itself, the Dockerfile pin policy block, the `_verdict_applied` double-count guard, the `cooldown` order fix in `_apply_verdict`, the lane-spill reasoning (consistent with the documented contract), repo-rule adherence (no `.env` writes, plans.example.yaml carries the new setting, docker-compose.yml not duplicating, tests placed before the `__main__` runner, no real-provider tests), or the CONCURRENCY-classification regex anchor.

tests/test_routing.py
Fledgewing: Fixed in commit a470a3f. Two changes:

1. Rewrote the docstring (tests/test_routing.py:1739) to state the actual invariant: backstop is an UPPER BOUND on the picker admit, and the picker admit is min(plan_cap, model.max_parallel) - what slots.try_claim enforces atomically. The previous assertion (backstop == cap_for(model, settings)) was too strong and missed the divergence.

2. The assertion loop now computes picker_admit per-model as min(policy.effective(plan).cap, model.max_parallel or ∞) and asserts backstop >= picker_admit_for_model. The example config still passes (every plan's backstop equals admit at the seed), but the test would catch a future config where the learner pulls effective() below the headroom-reduced backstop.

3. Added test_backstop_is_an_upper_bound_under_learner_pull_down: a synthetic 4-slot CLI plan with a model whose own max_parallel is None (so the headroom floor, not a model cap, does the narrowing), two consecutive connection refusals halve the learner to 2, backstop stays at 3 (headroom ceiling), picker admit drops to 2. backstop >= picker admit is verified end-to-end.

Suite green: all 40 routing tests pass, including the new divergence test.

switchyard/selfcheck.py
Fledgewing: Fixed in commit a470a3f. The fix in switchyard/selfcheck.py:_dynamic_probe passes the instance (ok_probe, rl_probe) to remove_callback_from_all_lists rather than probe.__class__, and drops the router reference in the finally block so the deployment-callback it adds is GC'd alongside the probe.

Verified the leak was real pre-fix and is closed post-fix:
- pre-fix: after selfcheck, litellm._async_success_callback / _async_failure_callback both still held the _ProbeImpl instance
- post-fix: both lists are empty after selfcheck

Reasoning on the manager's lookup (for the next reader): remove_callback_from_all_lists delegates to remove_callback_from_list_by_object which matches either by identity (c == obj, default require_self=False) or by bound-method self (c.__self__ == obj, require_self=True). Either path is identity-based. The class lookup key never matches because the stored callbacks are CustomLogger *instances* (not bound methods, so c.__self__ is undefined for them); the identity path checks c == probe.__class__ which is False because c is an instance and not its own class. Net effect: the previous cleanup was a no-op in this code path. The leak is harmless in the production entrypoint (selfcheck runs in a separate process before exec litellm) but bites any future caller that imports selfcheck in-process with a real litellm stack.

Suite green: full selfcheck PASSED on the sandbox litellm 1.102.0.

switchyard/selfcheck.py
Fledgewing: Fixed in commit a470a3f (switchyard/selfcheck.py:_audit_proxy_pre_call_hook_present).

Replaced the substring scan with an AST walk: parse each proxy .py, walk every ast.Call, and check whether 'async_pre_call_hook' appears in the callee's identifier chain (bare Name, any Attribute on the chain, or a Name at the bottom of a subscript). A bare 'async_pre_call_hook' identifier in a docstring or comment would not produce an ast.Call, so the gate now matches the loud semantic the contract is pinning to.

On the sandbox litellm 1.102.0 the AST walk finds 5 calls to async_pre_call_hook in the proxy package (proxy/utils.py:1571, proxy/utils.py:2222, proxy/policy_engine/pipeline_executor.py:501, plus two more the audit doesn't enumerate), so the static audit line stays green. A future litellm that renames the dispatch symbol or removes the call site entirely will trip the CRITICAL fail-loud gate.

Skipping SyntaxError when a proxy file fails to parse (forward-compat for a future half-broken litellm). The 'found_any = False' path then catches us on the no-calls outcome.

Fledgewing:
Replying to https://github.com/Fledgewing/SwitchYard/pull/60#issuecomment-5777321873 — both stale-reference items were fixed in commit a470a3f.

1. switchyard/picker.py:157 — the comment on the exclude= branch was rewritten to describe the actual contract now: the exclude parameter exists for the caller's retry re-entering through async_pre_call_hook (a previously-broken ref can be passed so the retry lands on a peer), with the explicit note that with num_retries=0 the spill IS the caller's retry, not a router-level re-pick. Also noted that tests drive the parameter directly for LaneSaturated paths. Kept the explanation of why refs rather than plans.

   Did NOT drop the parameter or the dead branch: tests/test_routing.py has 3 callers using exclude= (test_pick_honours_exclusions, test_pick_raises_lane_saturated_when_every_member_is_excluded, test_repick_moves_the_session_lease_and_returns_not_sticky) and removing it would also be a behaviour change beyond a documentation nit.

2. switchyard/portal/app.py:443 — the per-model cap cell in collect_plans() now passes reg.settings through to plan.cap_for(m, reg.settings), so CLI-backed plans get the same headroom-reduced cap that the litellm per-deployment backstop and the live picker admit both use. Without it, the row's cap could disagree with the lane's capacity block above by exactly the headroom value (gated by m.narrowed on the example config, but the conceptual gap was real).

tests/test_routing.py
Fledgewing: **Verified.** Re-read the rewritten docstring + assertion loop at HEAD a470a3f and re-ran the suite. Two things:

1. The docstring (lines 1739-1770 in the new file) now states the actual invariant: `backstop >= picker_admit_for_model` where `picker_admit = min(policy.effective().cap, model.max_parallel or effective_cap)`. That's exactly what `slots.try_claim` enforces atomically with the plan cap and the model cap as separate args, so the test now mirrors production.

2. The new `test_backstop_is_an_upper_bound_under_learner_pull_down` (lines 1873-1962) constructs a synthetic 4-slot CLI plan whose model has no `max_parallel` (so the headroom floor does the narrowing — the example plans all narrow at model level and would never trip the divergence), pulls the learner to 2 via two `note_rejection`s, and asserts `picker_admit < backstop` and `backstop >= picker_admit`. Re-ran it locally: `backstop=3 (headroom=1), picker admit=2 (learned[global])` — the divergence is exercised and the invariant holds. The original `test_the_generated_backstop_matches_what_the_picker_admits` still passes for the fixture plans (every backstop equals admit at the seed), which is exactly what the docstring now says: coincidence on the example, not a tautology.

Closing this as resolved.

switchyard/selfcheck.py
Fledgewing: **Partially fixed; carryover nit open.** Verified at HEAD a470a3f:

- The ProbeImpl leak is closed. Re-ran the probe in-process and dumped `litellm._async_success_callback` and `litellm._async_failure_callback`: both lists are empty of `_ProbeImpl` instances post-probe. Passing the instance instead of the class works (the manager's identity-based lookup matches `c.__self__ == obj` or `c == obj`).

- The Router leak that the original finding acknowledged is NOT closed, despite the comment at lines 425-436 claiming `ok_router = None` lets it be GC'd alongside the probe. The bound methods `Router.deployment_callback_on_success` and `Router.async_deployment_callback_on_failure` are added to `litellm._async_success_callback` / `_async_failure_callback` by `Router.__init__` (router.py:1089-1101), and the lists still hold them after the probe — verified by the same in-process dump. Bound methods keep their `self` alive, so the Router is reachable from the global state and stays alive past the `finally` block. The leak is harmless in the production entrypoint (separate process) and harmless when selfcheck is the only consumer, but for any future in-process caller that wants the selfcheck alongside a real router the Routers pile up.

  Suggested follow-up: also call `litellm.logging_callback_manager.remove_callback_from_all_lists(ok_router)` in each `finally` — that one extra call uses the same identity match to drop the bound methods and lets the Router be GC'd. The current comment should also stop claiming GC happens by reference drop alone.

Keeping the thread open for that one follow-up; will close it once the Router teardown is in.

switchyard/selfcheck.py
Fledgewing: **Verified for the case it was guarding, with one docstring-vs-impl nit open.**

Re-ran the audit on the sandbox litellm 1.102.0 — the AST walk now finds 5 dispatch sites (proxy/utils.py:1571, proxy/utils.py:2222, proxy/policy_engine/pipeline_executor.py:501, proxy/hooks/parallel_request_limiter_v3.py:3440, proxy/guardrails/guardrail_hooks/mcp_jwt_signer/mcp_jwt_signer.py:950). The bare-Name, attribute-chain, and chained-attribute cases all match. A docstring/comment-only mention of the identifier would no longer produce an `ast.Call` and would no longer satisfy the audit, which is the loud semantic the contract was pinning to.

One nit on the rewrite itself (which I'll keep on this thread rather than open a new one): the function-level comment at lines 241-244 promises `\`a[0](...)\` -> ['a']`, but the implementation returns `[]` for subscript-based callees — the while loop only chases `ast.Attribute`, and a `Subscript` is neither `Attribute` nor `Name` so the if-check after it doesn't fire either. The audit at lines 223-228 promises "a bare Name, an Attribute on a registered manager, or a subscript", so a future litellm that dispatches the hook via subscript (e.g. `hooks["async_pre_call_hook"](...)`) would silently fail the audit even though the dispatch still exists. Either:

- chase the `Subscript.value` once to collect the base name, or
- drop "or a subscript" from the high-level docstring and the function-level example so they agree.

The litellm source on disk doesn't currently use subscript dispatch, so this is a forward-compat gap rather than an active false-negative. Calling it out so the next reader doesn't trust the docstring.

Fledgewing:
Round-2 verification on the two stale-reference items (https://github.com/Fledgewing/SwitchYard/pull/60#issuecomment-5777321873):

1. `switchyard/picker.py:157-166` — comment rewrite now describes the caller-retry path through `async_pre_call_hook`, acknowledges that `exclude` is currently used only by tests (3 test files: test_routing.py:1012/1021/1059/1092), and explains the `(With num_retries=0 there is no router-level retry to feed this; the spill IS the caller's retry.)` clause. The forward-looking "exists for the caller's retry re-entering" is paired with the honest "Tests also use the parameter directly" so the next reader isn't misled. No production caller currently passes `exclude`, which matches the comment. **Resolved.**

2. `switchyard/portal/app.py:443` — now passes `reg.settings` so `plan.cap_for(m, reg.settings)` applies headroom to CLI-backed plans (lines 443-450). For Claude Max opus (no model cap), the row cap is now `max(1, 2-1)=1` matching the row's `capacity` block a few lines above (which already uses `policy.effective`). For fable (model cap 1, CLI-backed), both formulas collapse to 1. For API plans, no behavior change. Re-ran `tests/test_portal.py` — green. **Resolved.**

Both items close.

tests/test_routing.py
```diff
@@ -1685,6 +1692,276 @@ def test_perishable_one_adjacent_swap_reconciles_added_and_removed_refs():
...
+    reg = models.Registry(settings=reg.settings, plans=plans, lanes=reg.lanes)
+    redis = FakeRedis()
+    slots = SlotTable(redis, reg.settings.inflight_max_age_seconds)
+    ledger = Ledger(redis)
+    policy = CapacityPolicy(redis, reg.settings, ledger)
+    picker = Picker(reg, slots, policy)
+
+    async def go():
+        # Two consecutive refusals at the current concurrency halve the
+        # learned cap: 4 -> 2 -> 1; an additive probe-up would still be
```
Fledgewing: **[nit] `4 -> 2 -> 1` doesn't match the test's two rejections at concurrency 4**

The comment claims "Two consecutive refusals at the current concurrency halve the learned cap: 4 -> 2 -> 1", but the two `note_rejection` calls right below both pass `at_concurrency=4`. Tracing `ConcurrencyLearner.note_rejection`:

  new = max(1, int(at_concurrency * 0.5)) = max(1, 2) = 2
  first call: current is None, stored = 2, returns 2
  second call: current = 2 (from the previous hset), stored = min(new=2, current=2) = 2, returns 2

So the learner goes 4 -> 2 -> 2, not 4 -> 2 -> 1. The `2` floor that the test then asserts (`learned == 2` in the print line) holds, and the test passes — but the `-> 1` part of the comment is wrong, and a future reader who follows the math the comment suggests will not see what they were promised.

Either:
- drop the `-> 1` and say "Two consecutive refusals at concurrency 4 leave the learned cap at 2 (the floor of `int(4 * 0.5)` applied twice)", or
- actually drive the second halving by calling `note_rejection(plan, at_concurrency=2)` for the second rejection, which would land the learner on 1 and exercise the third halving the comment describes.

The test still validates the invariant (`backstop >= picker admit` and `picker_admit < backstop` both fire), so this is a docstring/expected-behaviour mismatch rather than a logic bug.

Fledgewing:
Round-2 review at HEAD `a470a3f` (commit a470a3fb4c72a9b156d02ca1cd7e430112bda669). Full offline suite green (40 routing, 16 classify, 27 policy); render_preview green; selfcheck PASSED on sandbox litellm 1.102.0.

**Round-1 fix verification** (replies on each thread):
- ✅ Backstop/picker invariant ([thread](https://github.com/Fledgewing/SwitchYard/pull/60#discussion_r4072097456)) — docstring rewritten to state the upper-bound invariant, assertion loop checks `backstop >= picker_admit_for_model`, new `test_backstop_is_an_upper_bound_under_learner_pull_down` exercises the divergence with a synthetic 4-slot CLI plan (verified locally: `backstop=3, picker admit=2`).
- ⚠️ Probe callback leak ([thread](https://github.com/Fledgewing/SwitchYard/pull/60#discussion_r4072105988)) — ProbeImpl leak closed; Router leak acknowledged but not fixed. The comment claims `ok_router = None` lets the Router be GC'd, but the bound methods the Router added to `litellm._async_success_callback` / `_async_failure_callback` (router.py:1089-1101) keep it reachable. Thread stays open for the one extra `remove_callback_from_all_lists(ok_router)` call + comment fix.
- ✅ AST walk ([thread](https://github.com/Fledgewing/SwitchYard/pull/60#discussion_r4072100536)) — finds 5 dispatch sites on litellm 1.102.0 (proxy/utils.py:1571, 2222; policy_engine/pipeline_executor.py:501; hooks/parallel_request_limiter_v3.py:3440; guardrails/.../mcp_jwt_signer.py:950). A small docstring-vs-impl nit kept on the thread (`a[0](...)` is documented to return `['a']` but actually returns `[]`; high-level docstring promises "or a subscript" that the impl doesn't deliver).
- ✅ Picker.py stale comment ([thread](https://github.com/Fledgewing/SwitchYard/pull/60#discussion_r4072100536)) — comment now describes the caller-retry path through `async_pre_call_hook`, explicitly notes "Tests also use the parameter".
- ✅ Portal `cap_for` with settings ([general comment](https://github.com/Fledgewing/SwitchYard/pull/60#issuecomment-5779418646)) — `app.py:450` now passes `reg.settings`; portal tests green.

**New findings posted in round 2:**
- [tests/test_routing.py:1929](https://github.com/Fledgewing/SwitchYard/pull/60#discussion_r4073497472) — nit: comment claims `4 -> 2 -> 1` but the test only fires `note_rejection(at_concurrency=4)` twice, and the learner's `current = min(new, current)` keeps the cap at 2, not 1. The `learned == 2` floor that the test relies on is correct; the `-> 1` portion of the comment is wrong.

Round-1 thread on the probe leak carries forward as a carryover nit (Router GC claim). The new docstring nit on the AST walk is open on its existing thread. Nothing new under the headroom / pin / CONCURRENCY-classification / repo-rules focus areas.

tests/test_routing.py
Fledgewing: Fixed in commit 254f767. Chose option 1 (drop the '-> 1' from the comment, leave the two note_rejection(at_concurrency=4) calls alone): the test's intent (two refusals to demonstrate the learner pulling down) is preserved, the picker_admit==2 invariant still holds, and the comment now describes what actually runs.

The new comment (tests/test_routing.py:1928) walks the math the reviewer laid out:
  new   = max(1, int(at_concurrency * 0.5)) = max(1, 2) = 2
  first call:  current=None, stored=2
  second call: current=2, stored=min(new=2, current=2)=2

So the descent is 4 -> 2 -> 2 (the second halving at the same at_concurrency is a no-op via the min(new, current) guard in switchyard/policy.py:note_rejection). The test prints backstop=3 (headroom ceiling), picker_admit=2 (learned[global]); the backstop >= picker admit invariant fires, and the test's contract is intact.

Suite green (no FAIL lines); test_routing.py 40 unchanged, test_classify.py 16 unchanged, test_policy.py 27 unchanged; selfcheck PASSED on sandbox litellm 1.102.0.

switchyard/selfcheck.py
Fledgewing: Fixed in commit 254f767. Two pieces, both in switchyard/selfcheck.py:

1. The Router bound methods (Router.deployment_callback_on_success and Router.async_deployment_callback_on_failure, registered by Router.__init__ at router.py:392-405 into litellm._async_success_callback and _async_failure_callback) are now explicitly torn down with require_self=True. The default require_self=False in remove_callback_from_list_by_object uses 'c == obj', which misses bound methods (a bound method is not equal to its self). require_self=True switches to 'c.__self__ == obj', which matches the bound method's self. Applied in both the 200-case and 429-case finally blocks; the 429 case also has a None-guard on rl_router for the case where Router.__init__ raises before the try block runs (mirroring how the 200 case protects ok_router by assigning it inside the try).

   Verified pre-fix: litellm._async_success_callback and litellm._async_failure_callback both held bound methods of the Router post-probe. Verified post-fix: both lists are empty after selfcheck.

   Safety reasoning: the probe has already completed by the time the finally block runs, so removing the Router from the callback lists only affects dispatch that would have happened on later requests inside this process. There are no later requests in either path (200 returns immediately, 429 raises and is caught and counted), so the removal is observably a no-op on probe outcomes; it only stops the leak. The production entrypoint runs selfcheck in a separate process before exec litellm, so the leak was harmless there; the fix is for any future in-process caller that wants the selfcheck alongside a real router.

2. The 200-case comment was rewritten to describe the actual three-way teardown (probe instance, Router bound methods with require_self=True, local Router reference) instead of claiming GC happens by reference drop alone.

   Also: I noticed the reviewer's suggested fix said remove_callback_from_all_lists(ok_router) and 'that one extra call uses the same identity match to drop the bound methods'. The default identity match (c == obj) does NOT drop bound methods - I had to pass require_self=True explicitly to get the c.__self__ == obj path. Flagging this so the next reader knows why the explicit require_self=True is there.

AST-walk docstring alignment: the docstring on _callee_names claimed 'a[0](...) -> [a]', but the implementation does not chase Subscript: when node.func is ast.Subscript, the while-loop does not enter (Subscript is not Attribute), and the final isinstance(cur, ast.Name) check fails, so the function returns []. Rewrote the docstring to state that explicitly: bare Names and Attribute chains are chased, subscripts (and calls-of-calls, etc.) are not. The pinned litellm dispatches async_pre_call_hook via attribute chains ('a.b.async_pre_call_hook(target)'), which the loop handles. A future build that dispatches via a subscript would need a smarter walker, but the audit's contract (prove the dispatch symbol is reachable by name on at least one call site) is what it is.

Suite green (no FAIL lines); test_routing.py 40 unchanged, test_classify.py 16 unchanged, test_policy.py 27 unchanged; selfcheck PASSED on sandbox litellm 1.102.0.

tests/test_routing.py
Fledgewing: **Verified.** Comment at lines 1928-1935 now states exactly what runs: "Two consecutive refusals at concurrency 4 leave the learned cap at 2. The learner applies min(new, current) on each call (policy.py:note_rejection), so the second halving at the same at_concurrency is a no-op: new=int(4*0.5)=2 and current=2 already, so stored stays 2." Re-ran with three consecutive refusals at concurrency 4 against a synthetic 4-slot CLI plan — the learned cap stays at 2 across all three calls (learner returns 2, 2, 2; effective cap=2 learned[global]), and the assertions (`backstop=3, picker admit=2`) still hold.

One trivial nit I am NOT posting as a blocker: the code now has THREE `note_rejection` calls at concurrency 4, while the comment talks about "Two consecutive refusals". The third is a no-op via the same `min(new, current)` guard the comment describes, so the test still does what the comment says — just with a redundant third call that proves the guard is idempotent. Worth tidying (drop the third call, or rewrite the comment to mention three) but not a block.

Closing this thread.

switchyard/selfcheck.py
Fledgewing: **Verified, fully closed.** Re-ran `python3 -m switchyard.selfcheck` in-process at HEAD `254f767` and dumped `litellm._async_success_callback` and `litellm._async_failure_callback` after the call — both lists are `[]`. The Router bound methods (`Router.deployment_callback_on_success`, `Router.async_deployment_callback_on_failure`) are no longer reachable from the global state.

The `require_self=True` switch on the second `remove_callback_from_all_lists` call is the right lookup path: `c == obj` (the default `require_self=False`) misses bound methods because a bound method is never `==` to its `__self__`; `c.__self__ == obj` matches. The three-way teardown (probe instance, Router bound methods, local `ok_router = None`) is correctly described in the comment at lines 449-467, and the 429 case's `if rl_router is not None:` guard covers the `Router.__init__` raising path before assignment.

Comment 436-437 ("pass the instance rather than the class. The manager's default lookup is `c == obj` (require_self=False), which matches the stored CustomLogger instance.") is also accurate.

Closing this thread.

switchyard/selfcheck.py
Fledgewing: **Verified for the function-level comment.** The new docstring at lines 241-254 is correct and aligned with the implementation: bare Name returns `['foo']`, Attribute chain returns `['a', 'b', 'c']`, subscripts (`a[0](...)`) and calls (`f()(...)`) return `[]`. Audited on litellm 1.102.0 — 5 dispatch sites still match.

One trivial consistency nit I am NOT posting as a blocker: the higher-level audit docstring at lines 223-228 still says "whether as a bare Name, an Attribute on a registered manager, or a subscript", but the function-level comment now says subscripts are not chased (return []). So the two docstrings in the same function disagree — the implementation is right (and the function-level comment matches it), but the higher-level one over-promises. Forward-compat concern only (the pinned litellm dispatches via attribute chains), and a one-line edit ("or a subscript" → drop the phrase) fixes it.

Closing this thread.

Fledgewing:
Round-3 review at HEAD `254f767` (commit 254f767d017913b0995455c9840d2f5c285e308a). Full offline suite green (40 routing / 16 classify / 27 policy); render_preview green; selfcheck PASSED on sandbox litellm 1.102.0; `python3 -m switchyard.gen_litellm config/plans.example.yaml /tmp/opencode/int-gen.yaml` emits `num_retries: 0`, `fallbacks: []`, `context_window_fallbacks: []`, `disable_cooldowns: true`, `max_parallel_requests: 1` for claude-max/opus (headroom-reduced).

**Round-2 fix verification** (replies on each thread):
- ✅ Test comment [4073497472](https://github.com/Fledgewing/SwitchYard/pull/60#discussion_r4073497472): now correctly describes the 4 -> 2 -> 2 descent via `min(new, current)`; re-ran with three consecutive refusals at concurrency 4 — learner returns 2, 2, 2; effective cap=2; assertions hold.
- ✅ Probe callback leak [4073486468 / 4072105988](https://github.com/Fledgewing/SwitchYard/pull/60#discussion_r4073486468): Router teardown now uses `require_self=True`; verified in-process that `litellm._async_success_callback` and `litellm._async_failure_callback` are both `[]` after selfcheck. The 429 case's `if rl_router is not None:` guard handles the `Router.__init__` raising path. Comment at lines 449-467 correctly describes the three-way teardown.
- ✅ AST walk function-level comment [4073489419 / 4072100536](https://github.com/Fledgewing/SwitchYard/pull/60#discussion_r4073489419): new docstring correctly says bare Name + Attribute chain chased, subscripts/calls return `[]`. 5 dispatch sites still match on litellm 1.102.0.

**Fresh-eyes pass at this HEAD — no new findings.** The contract this PR is landing (caller-retry spill with `num_retries=0`, `_verdict_applied` double-count guard, litellm tag+digest pin, startup self-check, `sidecar at capacity` CONCURRENCY classification, `gate_headroom_slots` floor 1 with `cap_for(model, settings)` alignment in `gen_litellm.py`) is implemented as described and verified end-to-end on the example config and the synthetic 4-slot CLI divergence case. Repo rules respected (no `.env` writes, plans.example.yaml carries the new setting, no docker-compose duplication of plans.yaml values, tests placed before `__main__` runner, no real-provider tests). One residual docstring inconsistency between the higher-level audit docstring (still says "or a subscript") and the now-correct function-level comment is forward-compat only and would be a one-line edit; not a block.

Issue #43, opened 2026-09-22
Lanes: nestable balancing strategies inside `order` (round-robin / weighted / headroom / perishable); `tail` becomes the final stage


Fledgewing:
## The problem

A lane order is flat and has exactly one behaviour: spill-and-fill down the list (plus the opt-in lane-wide `strategy: perishable` from #40). Two things you cannot say today:

- **"Spread across my twin seats instead of filling the first one."** `minimax-ultra/m3` and `minimax-max/m3` are identical $20 plans; ordered fill keeps the first seat pinned at its connection limit while the second idles, then cliff-edges over. The right shape is round-robin (or weighted) *between those two only*, then strict spill to `grok` and below.
- **"Balance here, drain-by-score there, all in one lane."** `strategy:` is lane-wide, so #40 forces all-or-nothing: either the whole body is perishable-ranked or none of it is. There is no way to perishable-rank the two Minimax seats while keeping `grok` strictly behind them.

## Proposed: order entries become *nodes* — a ref, or a strategy group

`order` stays a list. Each entry is either:

- a bare `plan/model` ref — today's behaviour, unchanged; or
- a **group**: the strategy name as the key, its members as the value. One key shape for every strategy — no `load_balance_round_robin:` family of long names. A group's value is a plain list of nodes, or a mapping (`members:` + options) when the strategy needs options.

A group member may itself be a bare ref **or another group** — nesting falls out of the recursion for free.

```yaml
forge:
  label: "Coding Workhorse"
  order:
    # the twin seats share new sessions instead of filling ultra first
    - round_robin:
        - minimax-ultra/m3
        - minimax-max/m3
    # or: spread them 5:2
    - weighted:
        - minimax-ultra/m3: 5
        - minimax-max/m3: 2
    # or: #40's score, scoped to just these members
    - perishable:
        - minimax-ultra/m3
        - minimax-max/m3
    - grok/grok-4.6
    - opencode-go/mimo-v2.6-flash
    - openrouter/mimo
  tail: [local-box/qwen]
```

Nesting (outer rotation picks a child; the child group resolves its own pick):

```yaml
order:
  - round_robin:
      - lowest_utilization:
          - minimax-ultra/m3
          - minimax-max/m3
      - grok/grok-4.6
  - opencode-go/mimo-v2.6-flash
  - openrouter/mimo
tail: [local-box/qwen]
```

### What a group *means*

[...]

PR #61, opened 2026-09-22
Lanes: nestable balancing strategies inside order (#43)

Fledgewing:
## Summary

Lanes now support nestable balancing strategies inside `Lane.order`. Each body
entry is either a bare `plan/model` ref OR a single-key `Group(strategy, members,
weights?)` — and group members may themselves be bare refs or further groups,
up to depth 4 (refused at parse). Five strategies ship: `fill` (default, today's
spill-and-fill, parses bit-for-bit identically), `round_robin` (alternates
members per session), `weighted` (visits by integer-weight ratio),
`lowest_utilization` (lighter-loaded member first; `perishable` minus the
hours-to-reset signal), and `perishable` (headroom-aware, drains emptiest
hour-per-dollar twin first). The unifying rule: a strategy produces a visit
order on each new-session placement; ordinary spill-and-fill runs through that
order, so affinity, per-member gates (5h, cooldown, spent-quota, supports_tools)
and the drain-promotion rule all keep working unchanged.

Tail stays a lane-level key, modelled as the final stage; pacing's
`disable_tail` survives (`tail disabled (pacing)` chip + log-line token both
present).

### How to verify

The offline test suite is the regression bar:

```
for t in tests/test_*.py; do python3 "$t" >/dev/null || echo "FAIL $t"; done
python3 tests/test_routing.py
python3 tests/test_portal.py
python3 tests/test_hot_reload.py
python3 tests/test_policy.py
```

Manual board render: open `/tmp/switchyard-preview.html` (from
`tests/render_preview.py`); the `forge` lane should show three group header
rows (`round_robin` / `weighted` / `perishable`) interleaved with three bare
plan/model rows, and `nest-demo` should show nested indent under its outer
`round_robin`. Smoke.py greps `lane=<lane> ->` and the `tail disabled (pacing)`
chip still match the gateway log line and the capacity board respectively.

### References

Issue **#43** — *Lanes: nestable balancing strategies inside order*.
Workstreams:
- **WS1** — `switchyard/{hooks,models,picker,usage}.py`, `tests/{fake_redis,test_routing}.py`
- **WS2** — `switchyard/portal/{app,groups}.py`, `switchyard/portal/templates/_capacity*.html`, `tests/{render_preview,test_portal}.py`
- **WS3** — `config/plans.example.yaml`, `README.md`, `TESTING.md`


switchyard/picker.py
```diff
@@ -142,138 +162,427 @@ async def _members(self, lane: str, needs_tools: bool = False) -> list[Model]:
...
+            return None
+        # 5h gate: a perishable or lowest_utilization group may have gated this
+        # ref on the 5h constraint window. The flag is read from the SAME hash
+        # the ranking lives in, so a missing/stale hash means "no gate", which
+        # is the right default — the picker falls back to config order. The
+        # `picked_group` here is the OUTERMOST group, not the one currently
+        # being walked: a ref reached via a nested group still wants the
+        # outermost scored group's gate, because the writer publishes one
+        # gate per leaf ref, not per nested member.
+        if picked_group is not None and picked_group.strategy in (
```
Fledgewing, line 255: **[blocker] gate5h is not enforced for refs nested inside a scored group when the outer group is not perishable/lowest_utilization**

The gate check at line 242 keys off `picked_group` (the OUTERMOST group), not the group whose hash actually holds the gate flag. So when a ref is reached via a nested perishable/lowest_utilization group whose OUTER group is `round_robin` (or any non-scored strategy), this branch never runs and the ref is served anyway.

The complementary eager-recording in `_visit_order_ranked` (lines 433-438) only adds `"<ref>(gate5h)"` to `ctx.skipped`. That makes the gate visible on the capacity board but does NOT skip the ref at the pick boundary.

Reproducer against the shipped `nest-demo` lane (which is outer round_robin over inner lowest_utilization): gate `claude-max/opus` and `claude-max/fable` in their respective inner rankings, drive four picks — every single one lands on the gated ref:

```python
# in test_routing.py
new_reg = _build_lane(reg, slots, key="nest-demo",  # already the shipped lane
                      order=[{"round_robin": [
                          {"lowest_utilization": ["claude-max/opus", "openai/sol"]},
                          {"lowest_utilization": ["claude-max/fable", "openai/astra"]},
                      ]}], tail=[], strategy="fill")
# Inner A: opus gate5h=1, sol gate5h=0. Inner B: fable gate5h=1, astra gate5h=0.
# 4 picks -> opus, fable, opus, fable (gated ones served every time)
```

The comment immediately above this block says "a ref reached via a nested group still wants the outermost scored group's gate, because the writer publishes one gate per leaf ref, not per nested member." That rationale is incorrect on its own terms: the per-group writer (`recompute_group_orders` -> `_collect_group_orders`) DOES recurse into nested groups and DOES publish one gate per leaf per nested group, so the data is in `sy:group-order:{inner_gid}:{lane}`. The picker just doesn't read that hash on the pick path.

The docstring on `_visit_order_ranked` (lines 397-401) says "The walk itself still skips gated members, so a stale hash that has since been refreshed cannot accidentally serve them." — but the walk does not skip gated members when the outer picked_group is round_robin. The docstring is wrong.

This violates the spec invariant at the top of issue #43: "Per-member gates. 5h gate, cooldown, spent-quota, supports_tools filtering, skip reasons must apply wherever a member sits. Strategies reorder; they never override a gate."

Suggested fix: instead of `picked_group` (the outermost), gate against the innermost scored group along the path. Either pass the deepest scored ancestor down through `_visit`, or have each `_visit_order_ranked` walk call short-circuit its OWN gated members (filter `ranked` to exclude nodes with `gate5h=True`) so the gate is enforced at the right level. Eager recording in `considered` is fine to keep; the walk just needs to actually skip.

A regression test pinning the nested case would also be welcome — `test_perishable_5h_gate_skips_a_member_for_a_new_session` only covers the lane-level sugar path and passes; the nested path is unexercised.

switchyard/models.py
```diff
@@ -494,6 +571,120 @@ def _parse_models(plan_key: str, raw: Any) -> dict[str, Model]:
...
+        if len(entry) != 1:
+            raise ValueError(
+                f"lane {lane_key!r} {path} must have exactly one strategy key, "
+                f"got {sorted(entry.keys())}")
+        strategy, value = next(iter(entry.items()))
+        if strategy not in _VALID_GROUP_STRATEGIES:
+            raise ValueError(
+                f"lane {lane_key!r} {path} uses unknown strategy {strategy!r}; "
+                f"must be one of {_VALID_GROUP_STRATEGIES}")
+        if depth >= _MAX_GROUP_DEPTH:
```
Fledgewing: **[should-fix] Group nesting depth limit is one tighter than documented (3 allowed, README says ≤ 4)**

`_MAX_GROUP_DEPTH = 4` and the check `if depth >= _MAX_GROUP_DEPTH` together reject depth=4, so the parser actually permits 3 nested group levels and rejects 4. The README and the references block both say "nesting depth ≤ 4" / "depth 4":

```
README.md:  Groups nest recursively up to depth 4, so an outer strategy can dispatch to inner strategies:
README.md:  covers the four named strategies, nesting depth ≤ 4, ...
```

and the commit message says "depth <= 4, refused at parse". An operator following those docs will hit the wall one level earlier than promised. The shipped `nest-demo` lane only goes 2 deep so the example does not show it, but a config that follows the docs literally (e.g. round_robin -> round_robin -> round_robin -> perishable) will fail to load with "exceeds max group nesting depth 4" — the constant in the message even implies the limit IS 4, which makes the off-by-one confusing.

Two clean fixes (pick one):
- `if depth > _MAX_GROUP_DEPTH:` (keeps the constant as the cap value), or
- bump `_MAX_GROUP_DEPTH = 5` so the current `>=` check accepts depths 1..4.

The boundary test (`test_nesting_depth_over_four_is_rejected`) does not pin this — it builds a 5-level tree and asserts rejection, which passes either way. Please tighten it to also assert that a 4-level tree loads and a 5-level one does not, so a future change doesn't silently tighten (or loosen) the limit.

Fledgewing:
Reviewed 2 findings: 1 blocker, 1 should-fix — see inline comments at 628d9d7.

The big one: the gate5h flag is only recorded on the capacity board (added to `ctx.skipped` eagerly) but is NOT actually enforced for a ref reached through a nested scored group when the outer group is round_robin/weighted — `_visit_ref` keys the gate check off the outermost `picked_group`, not the group whose hash carries the flag. Reproducer on the shipped `nest-demo` confirms: gating opus/fable in their inner rankings leaves every pick landing on the gated ref. The docstring at lines 397-401 ("the walk itself still skips gated members") is also wrong.

The smaller one: the parser rejects 4 levels of nesting while README, TESTING.md and the commit message all promise "≤ 4" / "depth 4"; the constant and the `>=` check disagree by one.

Everything else in the spec held up: affinity wins before the body walk, per-member gates (cooldown, supports_tools, exclude, spent-quota) apply inside groups, drain-promotion stays in its group, the gateway log line keeps the first token inside `[reason]`, the implicit `strategy: perishable` sugar still reads `sy:lane-order:{lane}`, the per-group writer populates `sy:group-order:{gid}:{lane}`, the integrator's three test fixes (routing:711 swap, flat-config snapshot, hot-reload node-level rotation) are all minimal and correct, and the 17-file stat matches the expected owned-files set.

PR #65, opened 2026-09-22
Portal: fix lane-card tag formatting — stack chips, drop configured+gate-headroom bubble, cap tag length

Fledgewing:
Status bubbles ("tags") on Lane cards stretched the card row when their text was long. This change, confined to `switchyard/portal/templates/_capacity_state.html`:

- Drops the diverged-reason chip **only** when the reason is the exact `"configured + gate headroom N"` shape — i.e. the chip is dropped only for the case where the learner-side reason was just `"configured"` and the sidecar's gate headroom was applied (the exact composition produced at `policy.py:423`). Learned (`learned[hour]` / `learned[global]`), paced (`paced N of M (...)`), and seed divergence reasons are still rendered: the `"+ gate headroom"` substring in the PR's first pass was too broad and over-dropped them.
- Wraps each status chip in a per-line `<div>` so bubbles stack vertically instead of flowing horizontally (follows the `_plans.html:147-153` alert-tags precedent).
- Truncates the visible diverged-reason text (`truncate(24)`) so bubbles stay short, with the full untruncated reason kept in the `title` tooltip (e.g. `title="{{ p.cap_reason }}"`).

The `cooled`, `failing`, `tail disabled (pacing)` and `days left` chips are byte-identical (tests assert them exactly). Offline suite and `tests/render_preview.py` green.


switchyard/portal/templates/_capacity_state.html
```diff
@@ -9,23 +9,36 @@
...
 {% else %}
-  <span class="muted">{{ p.in_flight }}/{{ p.cap }}</span>
-  {% if p.cap != p.cap_configured and not p.cap_model_owned|default(false) %}
-    <span class="tag warn" title="configured {{ p.cap_configured }}">{{ p.cap_reason }}</span>
+  <div style="margin-bottom:3px"><span class="muted">{{ p.in_flight }}/{{ p.cap }}</span></div>
+  {# Diverged chip: skip it when the only divergence is the sidecar's
+     gate-headroom reserve (policy.py appends "+ gate headroom N" to a
+     previously-"configured" reason). The withheld-slot tooltip on the
+     squares below the cap already carries the full reason. #}
+  {% if p.cap != p.cap_configured and not p.cap_model_owned|default(false) and "+ gate headroom" not in p.cap_reason %}
```
Fledgewing: **[should-fix] Jinja condition drops more chip cases than the PR description claims**

The added clause `"+ gate headroom" not in p.cap_reason` is a substring match, not a prefix/exact match against `"configured + gate headroom N"` as the PR description ("the 'configured + gate headroom N' combination") implies.

`policy.py:_apply_gate_headroom` (line 423) unconditionally appends `f"{cap.reason} + gate headroom {headroom}"` to whatever reason was set before, so in theory the chip reason can be:
- `"learned[h14] + gate headroom N"` / `"learned[global] + gate headroom N"` — when the learner is at or near configured and headroom still applies
- `"paced N of M (...) + gate headroom N"` — when the pacer keeps cap near `max_parallel` and headroom still applies
- `"configured (pacing: ...) + gate headroom N"` — when the pacer returned `None` and headroom still applies

In all those cases the chip carries pacing/learner info that the `title="configured {{ p.cap_configured }}"` attribute does NOT have — the title only says e.g. `"configured 4"`. Dropping the chip loses that info, contradicting the stated goal of dropping only the redundant "configured + gate headroom" case.

A safer check is a prefix/exact match on the simple case, e.g. `"configured + gate headroom " in p.cap_reason`, or — more robustly — gate the drop on the headroom flag exposed by `policy.effective()` rather than substring-matching the reason.

switchyard/portal/templates/_capacity_state.html
```diff
@@ -9,23 +9,36 @@
...
+    <div style="margin-bottom:3px"><span class="tag warn" title="configured {{ p.cap_configured }}">{{ p.cap_reason|truncate(24) }}</span></div>
```
Fledgewing: **[should-fix] Title attribute does not carry the full cap_reason; the "full value in the title tooltip" claim is wrong**

The chip's `title` attribute is `title="configured {{ p.cap_configured }}"`, e.g. `"configured 4"`. It has never contained the full `cap_reason`; this PR doesn't change that.

After the visible text is `truncate(24)`-shortened on this line, a user hovering it still sees `"configured 4"` — not the full reason that was previously in the visible text. The only way to recover the full reason is to hover the withheld-slot square below the cap (`_capacity_slots.html:61`, `title="withheld: {{ p.cap_reason }}"`).

So the PR description's "with the full value kept in the \`title\` tooltip" is incorrect, and the in-template comment ("the full reason still rides on the title attribute for the chips that have one") is misleading — the title never had it.

Either fix the description/comment to match what the chip actually does, or move the full reason into the chip's `title` attribute, e.g. `title="configured {{ p.cap_configured }} — {{ p.cap_reason }}"`, so the chip's hover affordance actually recovers what its truncation removed.

Note also that Jinja's `truncate` defaults to `leeway=5`, so strings of length 25-29 pass through unshortened — anything ≤ 29 chars is shown in full, not ≤ 24 as the description's "cap tag length" wording implies. Pass `leeway=0` if a hard 24-char cap is intended.

Fledgewing:
Reviewed 2 findings — both should-fix on the diverged-chip condition and the truncate/title claim. Byte-asserted chips in tests/test_portal.py (lines 181-182, 523) are untouched as the description claims, and `tests/test_portal.py` + `tests/render_preview.py` both run green locally. See inline comments at commit b898d31.

PR #66, opened 2026-09-22
Caller environment is authoritative: fix #44 (tabs see the relay's Linux env)

### LINKED ISSUE #44: Tabs on the gateway think they're on Linux in /app/mcp_bridge (client sends no environment, inner CLI adds its own)
Same reporter as #29, #30, #35 and #36. Thanks for merging #40 and #42 so fast.

## What happens

A real Claude Code tab pointed at the gateway (`ANTHROPIC_BASE_URL`) says hi, and then tells the user it is on **Linux, in `/app/mcp_bridge`**, and that their normal tools and drives "aren't reachable from here". None of that is true: the tools are proxied back and run on the user's machine. Asked directly, "what OS am I on and what folder are we in?" came back `Linux (WSL2, with the exact kernel version), working in /app/mcp_bridge` in 3 out of 3 tries. The user was on Windows.

## Root cause (measured, two halves)

1. **The client sends no environment.** I captured what Claude Code 2.1.278 sends to a custom base URL, using a local recorder in place of the gateway. The system prompt is about 6.2k chars and has **no environment section at all**: no working directory, no platform. The first user message carries CLAUDE.md and the reminders, but no environment either.
2. **The inner CLI adds its own.** Even with `--system-prompt`, the CLI inside the sidecar appends its own `# Environment` block (`Primary working directory: /app/mcp_bridge`, `Platform: linux`, `OS Version: <the WSL2 kernel>`). So that is the only "where am I" the model ever sees. Note that `--exclude-dynamic-system-prompt-sections` is ignored when `--system-prompt` is set.

Also: `mcp_bridge` never appends a sandbox note the way the text path does in `run_cli`. Every interactive tab carries tools, so every interactive tab takes the mcp path.

## What fixed it here

- **The system note alone is not enough.** With a "your environment describes the relay, not the user" note appended to `system` in `mcp_bridge.start_session`, the model *still* said Linux 3/3. It ranks the inner `# Environment` block above anything earlier in the system prompt. Adding the note is still worth doing, for path handling.
- **Give the model the real environment.** The launcher that sets `ANTHROPIC_BASE_URL` adds `--append-system-prompt` with the real working directory, platform and shell, plus a line saying that any Linux, `/app` or `/tmp` environment belongs to the relay. That text does reach the inner CLI; I confirmed it with a canary.
- **Put a one-line reminder at the top of the first user turn**, only when that client block is present, pointing at it. The prompt is the last thing the model reads before answering. It only goes on the first turn of a session.

With all three: 2/2 correct ("Windows, `C:\Users\<user>`"), "can you reach my drive?" got a correct yes with a live directory listing, and tool calls were unaffected.

## Suggestions

- Add the sandbox note on the mcp path, matching the text path.
- Consider documenting (or shipping) a client-side launcher snippet that appends the real environment. The gateway itself can't know the caller's machine, and the client doesn't send it.
- If you'd rather keep all of it server-side: a request header carrying the caller's cwd and platform, turned into the reminder by the bridge, would do the same without a launcher. That would need LiteLLM to forward the header to the sidecar.

Happy to share the diffs; they're small.

Fledgewing:
## Summary

Tabs pointed at the gateway answered `Linux, /app/mcp_bridge` because the
inner CLI's own `# Environment` block describes the relay container and
nothing told the model otherwise. This makes the caller's tool-execution
environment a first-class, per-session concept so the model is told the
real platform / cwd / shell instead of the relay's.

Fixes #44.

## Spec

Implemented per the OWNER (Fledgewing) comment on 2026-09-22 17:10:52
on issue #44:

- Per-session `CallerEnvironment` (cwd, platform, shell, source).
- Resolution precedence: **config > request > probe > host-platform > unknown**.
- The relay's environment is **never** substituted as a fallback: the
  sidecar is the sidecar, and saying otherwise is the bug.
- `caller_environment` block in `config/plans.example.yaml` documents the
  config-level override, the host-platform hint (`fallback_platform`,
  platform only), and the `probe: auto|disabled|required` mode switch.
- Self-describing `switchyard_env_<fingerprint>` probe ids: the id is the
  primary key, the fingerprint is recoverable across restarts, no
  in-memory pending dict is required. The probe mints a single
  read-only synthetic tool call on the very first request of a new
  session when nothing else has yielded an env, half a turn before the
  inner CLI session exists.
- Probe command is exactly `printf 'cwd=%s\nplatform=%s\nshell=%s\n'
  "$(pwd)" "$(uname -s)" "$SHELL"` — never `env` / `printenv`. Only
  enumerated shell variable is `$SHELL`. Never enumerates env vars.
- No sidecar session or gate slot is held while probing: the probe
  response is a tiny synthetic chat.completion that travels back to the
  caller, who executes it locally and answers on the next request. The
  next request is what builds a real session.
- Probe is one-shot: `FAILED_PROBES` and `RESOLVED_PROBES` caches are
  bounded by `MCP_PROBE_CACHE_LIMIT`; a refused probe is `isError` /
  unparseable → fallback env, no retry.
- Inner CLI cwd moved off `/app/mcp_bridge` (and `/app/cli_bridge`) to
  a per-session / per-call temp dir (`mcpb-<id>-...`, `sy-cli-...`).
  Cleaned up on every exit path.
- `[SwitchYard tool execution environment]` system block on both bridges
  (text path + mcp path), explaining that the relay's environment is
  the relay's.
- One-line `[SwitchYard: tools execute on ...]` first-real-turn
  reminder on the first user turn of the rebuilt / fresh session.
- Probe is recognised on shape, not name: tool schemas are inspected
  (Claude Bash-like, OpenCode shell, generic command, PowerShell-
  compatible). Name-based matching that is fragile to renames is
  rejected; only command-shaped tools with the right argument key are
  picked up. No `Bash`-name coupling.
- Passive parsers cover OpenCode (`<environment>` block), Codex
  (`# Environment` context), Anthropic (`<env>` block), and a generic
  first-turn `Working directory:` / `Platform:` / `Shell:` block.
- Relay-looking cwds (`/app/...`, `\\app\\...`) are dropped from any
  source: a cwd that looks like the relay is the relay's, not the
  caller's.
- Unknown-CLI / refused-probe paths emit **explicit "unknown"**
  wording, never the relay's Linux / `/app`.

## Acceptance evidence

Offline suite (run on this commit, `for t in tests/test_*.py; do python3 "$t" >/dev/null || echo "FAIL $t"; done`, all green):

- `tests/test_caller_env.py` — **31 tests passed**: parsers
  (OpenCode / Codex / Anthropic / generic first-turn), the
  precedence chain (`config` > `request` > `probe` > `host` >
  `unknown`), the probe-id round-trip (mint / parse, restart-safe,
  restart-stable fingerprint), the command-tool recogniser
  (Claude-Bash-like, OpenCode-shell, generic-command,
  PowerShell-compatible, plus rejections for non-command-shaped
  entries and `Bash`-with-bad-schema), the renderers (known env,
  unknown env, no host-env leakage), the relay-cwd drop, and the
  probe command being exactly `pwd` + `uname` + `$SHELL`.
- `tests/test_mcp_bridge.py` — **44 tests passed** (39 → 44): probe
  freshness (synthetic probe minted only on unknown env, no slot
  taken), probe consumption (synthetic exchange stripped, env
  parsed, system has env block), failed probe (no retry on second
  attempt, env rendered as `unknown`), metadata stamp honoured
  when present, passive env bypasses probe entirely, inner CLI
  spawned in per-session workdir (`mcpb-...`), rebuild path also
  resolves env and re-emits first-turn reminder.
- `tests/test_cli_bridge.py` — **44 tests passed** (38 → 44): text
  path spawn cwd is a fresh temp dir (not `/app`), text path argv
  carries env block when resolved, `SYSTEM_MODE=replace` keeps env
  block on `--system-prompt` and preserves
  `--exclude-dynamic-system-prompt-sections`, unknown env renders
  "unknown" wording with no Linux / `/app` leakage, no-tools +
  metadata-stamped `caller_env` does not 400 `tools_unsupported`.
- `python3 tests/render_preview.py` — green: rendered index +
  fragments (87899 bytes); 10 plans, 19 models, 5 lanes, 18 quota
  windows, 4 probes.

Coverage matrix vs the OWNER's acceptance list on issue #44:

| Acceptance item                                              | Unit-tested? |
| ------------------------------------------------------------ | ------------ |
| OpenCode on Windows / macOS → Linux SwitchYard               | yes (passive parser + env block) |
| Codex → Linux SwitchYard                                     | yes (passive parser + env block) |
| Claude Code (no env block) → SwitchYard                     | yes (probe mint + consume)       |
| Multiple simultaneous callers (Windows/macOS/Linux)         | yes (per-request resolution)     |
| Unknown CLI with usable shell tool → env bootstrap          | yes (tool-shape recogniser)      |
| Unknown CLI with no usable env data and no shell tool        | yes (`unknown()` wording)        |
| Probe rejected → no retry loop, fallback continues          | yes (FAILED_PROBES cache + no-retry) |
| Explicit config override wins                               | yes (`config` beats everything)  |
| Remote caller OS differs from Docker host → caller wins     | yes (precedence + host hint)    |
| Caller CWD differs from `docker compose up` cwd → caller wins | yes (host never substitutes a cwd) |
| Sidecar Linux never represented as caller environment       | yes (relay-cwd drop + unknown wording) |

The only items the offline suite cannot prove are the two manual checks
in TESTING.md § 8 (gateway metadata transport over LiteLLM, and a real
Claude Code tab against the live stack). Both need a running gateway.

## Manual post-merge steps for the operator

`switchyard/*.py` and the sidecars are baked into three images; the worktree
does **not** rebuild them. From the **main checkout**, after the PR is merged:

```
scripts/apply.sh --build
```

(rebuilds `gateway`, `portal`, `claude-max-sidecar`, `xai-token-proxy`, and
the `*-mcp-bridge` / `*-cli-bridge` sidecars with the new
`switchyard/caller_env.py`, the updated bridges, the `caller_environment`
config block, and the Dockerfile-sidecar import wiring).

Then run the two TESTING.md § 8 smoke checks against the live stack:

1. § 8a — metadata-transport assumption: send the same request twice,
   once without and once with `metadata.switchyard.caller_env`,
   confirm the sidecar honours the stamp and that the rendered CLI
   prompt carries the `[SwitchYard tool execution environment]` block
   with the right platform / cwd.
2. § 8b — real Claude Code tab: point a Claude Code tab at the gateway,
   open a session on each of Windows / macOS / Linux, ask
   `what OS am I on and what folder are we in?` — expect the caller's
   OS and the caller's folder, not `Linux (WSL2, ...)` and
   `/app/mcp_bridge`.

The offline suite asserts the resolution + rendering half of the
contract; only LiteLLM's runtime metadata transport and a real
client round-trip can prove the rest.


switchyard/models.py
```diff
@@ -150,6 +150,43 @@ class TransientBreaker:
...
+    sidecar is the sidecar, and saying otherwise is the bug.
+
+    `probe` controls whether the mcp_bridge may emit a synthetic
+    environment-discovery tool call when neither (1) nor (2) gave an
+    answer. `auto` is the right default for a permissive caller; a
+    kiosk with no callable shell tool would still leave (4) and (5).
+    `disabled` skips the probe entirely; `required` refuses requests
+    that have neither an explicit override nor a passive environment.
+    """
+    probe: str = "auto"             # auto | disabled | required
```
Fledgewing, line 183: **[blocker] `probe: required` is documented but never enforced**

The docstring above promises:

> `required` refuses requests that have neither an explicit override nor a passive environment.

In practice, `probe == "required"` is checked nowhere. The only consumer of `caller_environment.probe` is `mcp_bridge.server._maybe_probe` (`if getattr(ce_settings, "probe", "auto") == "disabled": return None`), which short-circuits only for `"disabled"`. When the operator sets `probe: required` and the env resolves to `unknown`, the code falls through to `handle_fresh` / `_handle_chat` and renders the explicit "unknown" wording — the exact behaviour the operator was trying to forbid.

The cli_bridge text path doesn't even check `probe` at all (it doesn't probe), so a `required` text request also gets through as `unknown` instead of 4xx.

`tests/test_caller_env.py` has 31 tests and `tests/test_mcp_bridge.py` has the 6 new probe tests, and none of them assert anything about `required`, so the gap is also unverified end-to-end.

Suggested fix: at the resolution layer (after passive parse, after the probe attempt, before `CallerEnvironment.unknown()` is returned to the renderer), check `cfg.probe == "required"` and raise an `HTTPException(status_code=400, detail={"error": {"type": "caller_environment_required", ...}})` so the sidecar refuses instead of rendering `unknown`. Mirror in both `handle_fresh` and `_handle_chat`. Add a test that sets `probe="required"` on a fixture with no body env and asserts the 4xx.

switchyard/caller_env.py
```diff
@@ -0,0 +1,582 @@
...
+        "relay container is the caller machine; ask if needed.]"
+    )
+
+
+# ---------------------------------------------------------------------------
+# Resolution
+# ---------------------------------------------------------------------------
+def resolve(data: dict[str, Any], cfg: Any,
+            probe_result: CallerEnvironment | None = None,
+            host_hint: CallerEnvironment | None = None) -> CallerEnvironment:
```
Fledgewing: **[nit] `host_hint` parameter is dead code**

`resolve()` accepts `host_hint: CallerEnvironment | None = None` but the body never references it. The host fallback is read from `cfg.fallback_platform` (lines 577-580), not from `host_hint`. The mcp_bridge callers (`handle_fresh` line 1241, `resume_gone_session` which doesn't pass it) always set `host_hint=None`.

The test `tests/test_caller_env.py:223` `test_resolve_host_fallback_only_platform_never_cwd` constructs a host_hint with a cwd and passes it in, then asserts the result has `cwd is None`. The assertion is correct, but it passes because of `cfg.fallback_platform`, not because the function does anything with `host_hint` — the test would pass identically with `host_hint` left at the default.

Either delete the parameter (and drop the `host_hint=...` argument at the two call sites) or actually wire it in (e.g. merge its platform into the host branch, with the cwd suppression already done by `_is_relay_path`). Leaving it in is misleading — readers will assume the docker-host's cwd is being threaded through.

sidecars/mcp_bridge/server.py
```diff
@@ -393,6 +444,179 @@ def session_id_from_call_id(call_id: str) -> str | None:
...
+def _resolve_env(body: dict, *, known: Any = None) -> Any:
+    """Stamped env first (if any of the trusted sources), else passive parse,
+    else unknown. Caller may pass `known` (e.g. from probe or cache) to skip
+    parsing and trust the value -- the contract is "I already know this"."""
+    if known is not None:
+        return known
+    if _caller_env is None:
+        return None
+    meta = (((body or {}).get("metadata") or {}).get("switchyard") or {}).get("caller_env")
+    if isinstance(meta, dict) and meta.get("source") in ("config", "request", "host", "probe"):
```
Fledgewing: **[should-fix] Caller-controlled `metadata.switchyard.caller_env` stamp is trusted unverified**

Both bridges read `body.metadata.switchyard.caller_env` and accept whatever `source` the caller claims, as long as it's in the trusted list (`("config", "request", "host", "probe")` here in mcp_bridge; `"probe"` excluded in cli_bridge's text path at line 1574). Nothing checks that the stamp was actually minted by the gateway.

Concretely:

```python
if isinstance(meta, dict) and meta.get("source") in ("config", "request", "host", "probe"):
    try:
        return _caller_env.CallerEnvironment(
            cwd=meta.get("cwd"), platform=meta.get("platform"),
            shell=meta.get("shell"),
            source=meta["source"] if meta.get("source") in
                   ("config", "request", "probe", "host", "unknown") else "unknown",
        )
```

A caller that can reach the sidecar directly (e.g. via a misconfigured compose network or an exposed port) can stamp `metadata.switchyard.caller_env = {"platform": "linux", "cwd": "/app/mcp_bridge", "shell": "/bin/sh", "source": "config"}`. The sidecar honors it and renders that into the system prompt verbatim, claiming the operator forced it — exactly the relay-env-as-caller-env bug the whole change is meant to prevent.

The mcp_bridge header comment says the gateway stamp is "best-effort" and "the sidecar must not require it", which is good. But "must not require" and "must trust" are not the same: the sidecar should *fall back to* the body parse, not *prefer* an unverified stamp.

Two viable fixes:

1. Always run `_caller_env.resolve(body, cfg)` first and use the stamp only as a fast-path override when its source is strictly stronger than what the body resolves to. (E.g. body resolves to `host`, stamp says `request` → use the stamp; stamp says `host`, body resolves to `request` → ignore the stamp.) This keeps the precedence honest.
2. Drop the metadata stamp path entirely on the sidecar and always re-resolve. The stamp stays on the wire as observability (logs, board) but never influences rendering. Simpler, and removes the attack surface entirely.

The same fix needs to land in `sidecars/cli_bridge/server.py:1574` (text path). Without it, anyone who can hit the cli_bridge directly with a stamped metadata field can also lie about `source="config"`.

sidecars/mcp_bridge/server.py
```diff
@@ -393,6 +444,179 @@ def session_id_from_call_id(call_id: str) -> str | None:
...
+    if ce_settings is None:
+        return None
+    if getattr(ce_settings, "probe", "auto") == "disabled":
+        return None
+    messages = body.get("messages") or []
+    env = _resolve_env(body)
+    if env is not None and env.source != "unknown":
+        return None
+    fp = _caller_env.fingerprint(messages)
+    if fp in FAILED_PROBES:
```
Fledgewing: **[should-fix] Probe-mint race: two concurrent requests with the same fingerprint can both mint probes**

`_maybe_probe` is the only place that gates probe minting, and the gate is not atomic with the mint:

```python
fp = _caller_env.fingerprint(messages)
if fp in FAILED_PROBES:           # check
    return None
if fp in RESOLVED_PROBES:         # check
    return None
tool = _caller_env.find_command_tool(tools)
if tool is None:
    return None
name, arg_key = tool
call_id = _caller_env.mint_probe_call_id(fp)   # mint
```

Two `asyncio.run(server.handle_tool_request(...))` calls that share a fingerprint (e.g. the same first-two-messages and same client retried in parallel, or a poorly-clocked caller that bursts two requests before either has its stamp set) both pass the `in FAILED_PROBES` / `in RESOLVED_PROBES` checks, both find a command tool, both mint `switchyard_env_<same fp>`, and both synthetic tool_calls go out. The caller gets duplicate probe tool calls in two separate HTTP responses; whichever result comes back first populates `RESOLVED_PROBES`, the second is silently dropped on `_consume_probe_results` (`remember_probe_env` overwrites) but the duplicate request still consumed a slot on the way to the response and the duplicate is in the caller's history.

Worse, both calls race into `handle_fresh` afterwards if their respective "first" probe replies fail — both mint, both fail, both `mark_probe_failed` with the same fingerprint, and the second one has nothing to overwrite. Mostly harmless but noisy.

The fix is small: guard the check+mint pair with an `asyncio.Lock` keyed on the fingerprint (or a single lock around both caches), and update `RESOLVED_PROBES` *inside* the lock so the second concurrent caller sees `fp in RESOLVED_PROBES` after the first one publishes. The probe cache is already bounded, so a lock doesn't add a memory concern.

`tests/test_mcp_bridge.py:test_fresh_with_unknown_env_emits_synthetic_probe_tool_call` covers the single-request case but doesn't exercise the race; a regression here would only surface under load.

Fledgewing:
Reviewed 4 findings: 1 blocker, 2 should-fix, 1 nit — see inline comments at 317bd9d.

Brief recap of the offline-suite evidence (`for t in tests/test_*.py; do python3 "$t" >/dev/null; done`):

- `tests/test_caller_env.py`: 31 passed
- `tests/test_mcp_bridge.py`: 44 passed (39 → 44)
- `tests/test_cli_bridge.py`: 44 passed (38 → 44)
- `tests/render_preview.py`: green (10 plans, 19 models, 5 lanes, 18 quota windows, 4 probes)

Test counts in the PR body match. All 31 / +6 / +5 are inserted before the `__main__` runners. No real provider is reached (`tests/conftest.py` blocks every non-loopback socket).

What's solid:

- Precedence chain in `caller_env.resolve` matches the OWNER's spec exactly.
- Relay-path suppression (`/app/...`, `/tmp/switchyard-relay/`, the new `mcpb-*` and `sy-cli-*` prefixes) is consistent across `parse_request`, `parse_probe_result`, and `fb_platform`.
- Probe id mint/parse round-trip is restart-safe (fingerprint recoverable from the id alone, no in-memory dict needed).
- `PROBE_COMMAND` is exactly `printf 'cwd=%s\nplatform=%s\nshell=%s\n' "$(pwd)" "$(uname -s)" "$SHELL"` — no `env`, no `printenv`, only `$SHELL` enumerated. Verified by `test_probe_command_contains_no_env_enumeration`.
- Synthetic probe exchange is stripped before the CLI sees it (`_consume_probe_results` + `_strip_synthetic_assistant`).
- Inner CLI cwd moved off `/app/{mcp_bridge,cli_bridge}` to per-session / per-call temp dirs and cleaned up on every exit path.
- `Dockerfile.sidecar` follows the `Dockerfile.token_proxy` precedent (`switchyard/__init__.py` + targeted `COPY`, `PYTHONPATH=/app`); `docker-compose.yml` untouched.
- `config/plans.example.yaml` carries the new `caller_environment` block; the live `plans.yaml` is gitignored and unaffected.

What I flagged:

- **Blocker**: `probe: required` is documented in `CallerEnvironmentSettings` but never enforced anywhere — both bridges treat it like `auto` and render the `unknown` wording. No test covers the mode.
- **Should-fix**: Both bridges trust `metadata.switchyard.caller_env.source` at face value (no gateway verification). A caller that reaches the sidecar directly can stamp `source="config"` and bypass the precedence chain — exactly the relay-env-as-caller-env bug the change is meant to prevent.
- **Should-fix**: `_maybe_probe` has a check-then-mint race on `FAILED_PROBES`/`RESOLVED_PROBES`. Two concurrent requests with the same fingerprint both mint duplicate probes.
- **Nit**: `host_hint` parameter to `caller_env.resolve()` is dead code — the fallback_platform comes from `cfg.fallback_platform`, not from `host_hint`. The test at `test_resolve_host_fallback_only_platform_never_cwd` only happens to pass because of that.

Verdict: comment-only review, no code edits. The pieces that the offline suite can verify (parsers, probe-id round-trip, system block rendering, per-call cwd, env-block argv wiring, no-tools passthrough, SYSTEM_MODE=replace preservation) are all in good shape. The blocker + should-fixes above need either code changes or a documented rationale before this is merge-ready.

sidecars/mcp_bridge/server.py
```diff
@@ -997,6 +1260,36 @@ async def handle_fresh(body: dict, tools: list[dict],
...
+            except _caller_env.CallerEnvironmentRequired as exc:
+                # The operator set probe=required and we cannot resolve
+                # the env. Refuse the request instead of rendering
+                # "Caller platform: unknown" -- the whole point of the
+                # option is loud refusal. 503 matches the file's existing
+                # capacity / resume-gone conventions, with a Retry-After
+                # so a caller whose CLI just gained permission can retry.
+                log.warning("probe=required but env unresolvable; refusing "
+                            "request: %s", exc)
+                raise HTTPException(
```
Fledgewing, line 1284: **[should-fix] `probe: required` refusal uses 503 + Retry-After: 5, but the failure mode is not transient**

The 503 + Retry-After pattern in the rest of this file (lines 1080 / 1530) is reserved for transient capacity issues — `acquire_waiting` failed within RESUME_WAIT_SECONDS, retry in 15s/30s and a slot may have freed. `probe=required` refusal is different: nothing about the request, the caller, or the operator's plans.yaml can change in 5 seconds. The env will still be unresolvable on retry for at least three of the four documented cases:

- Caller CLI still refuses the probe → retry yields the same refusal.
- Body still has no passive env → retry yields the same body.
- Operator's plans.yaml still has no fallback_platform and no forced values → retry yields the same resolve.
- (The only case where a 5s retry is meaningful is "operator just edited plans.yaml" — unlikely to be a steady-state.)

So `Retry-After: 5` actively misleads the caller. Either:

1. Use 400 (Bad Request) — semantically the request cannot be served because the operator's config requires an env that isn't there. The detail.type=caller_environment_required and the explanatory message already convey "this isn't transient; fix the request or the config." No Retry-After.
2. Or use 503 with a much longer Retry-After (e.g. 300s) to signal "this might resolve when an operator acts" without implying a 5-second turnaround.

The new test `test_probe_required_unresolvable_returns_503_no_unknown_wording` hard-codes both `status_code == 503` and `headers["Retry-After"] == "5"`, so flipping the convention means updating both the sidecar and the test.

Also worth noting: the same 503 + Retry-After: 5 is repeated verbatim in `resume_gone_session` (line 1513) for the rebuild path, so any fix lands in both call sites.

switchyard/caller_env.py
```diff
@@ -0,0 +1,653 @@
...
+    exactly the relay-env-as-caller-env bug the change exists to
+    prevent. The bridges translate this to an HTTPException (503 on
+    the mcp path; the text path downgrades `required` to `auto` before
+    calling resolve, since there is no probe possible there and
+    refusing every request that lacks a passive env would be the wrong
+    default for a permissive caller).
+    """
+
+
+def from_wire_metadata(meta: dict | None) -> CallerEnvironment | None:
```
Fledgewing, line 131: **[should-fix] `from_wire_metadata` is more permissive than the code it replaced**

The new helper returns a `CallerEnvironment` for ANY stamp with at least one field, including:

- `{"platform": "linux"}` → source="unknown", but the values flow through.
- `{"platform": "linux", "source": "made_up"}` → source="unknown" (the test even pins this case at line 351).

The old code (mcp_bridge/server.py round-1 lines 484-491, cli_bridge/server.py round-1 lines 1572-1581) only honored stamps whose source was in `("config", "request", "host", "probe")` — anything else (None, "unknown", "made_up") fell through to `parse_request(body)`. The new code returns parsed for those too, and both bridges then use it without consulting the body.

In the threat model where the sidecar is exposed to untrusted callers (the round-1 concern), this is a small but real expansion of the attack surface: an attacker who wants to inject `platform=linux`, `cwd=/app/mcp_bridge` no longer needs to bother including `source="config"` (now relabeled) or even knowing what the trusted-source list is — they can just send a stamp with values and no source. The values still win because the stamp is consulted before the body parse.

Two viable tightenings:

1. Add a "trusted source" gate inside `from_wire_metadata`: only return parsed when `raw_source in ("config", "request", "probe", "host")`. `None` and `"unknown"` and unknown labels return None, so the caller falls back to `parse_request(body)` — matching the pre-fix behavior. The test on line 351 would need to be updated to assert `from_wire_metadata({"source": "made_up", "platform": "linux"}) is None`.
2. Keep the current "values always win" semantics but document it explicitly in the docstring — and accept that source="unknown" stamps are trusted at face value. This is what the current code does; it's a deliberate choice, just one that wasn't called out in the PR description.

The test coverage gap is also worth noting: the existing tests exercise stamps with source="config"/"request"/"made_up", but there is no test for `{"platform": "linux"}` (no source) reaching `_resolve_env` on the mcp path. A future refactor that "fixes" the no-source case to fall through would have no test failure to catch it.

Fledgewing, line 131: **[should-fix] The metadata-spoof fix is partial: only `source=config` is re-labeled**

The fix targets source="config" specifically because only the operator's plans.yaml can legitimately produce that label. But `source="host"` is also derived from plans.yaml — it's the label `resolve()` returns for the `fallback_platform` branch (line 642). A caller who can reach the sidecar directly can stamp `metadata.switchyard.caller_env = {"platform": "linux", "cwd": "/app/mcp_bridge", "shell": "/bin/sh", "source": "host"}` and `from_wire_metadata` returns it with `source="host"` honored verbatim. The values then win because the stamp is consulted before the body parse.

Same exploit class as the round-1 config-spoof, just at a different tier. The relabeling (or rejection) needs to extend to `"host"` too, OR — the cleaner fix — the bridges should drop the metadata stamp path entirely and re-resolve from the body every time, treating the stamp as observability only.

(For what it's worth: the source="probe" stamp is also caller-controlled, but in practice the gateway cannot mint it — only the sidecar's `_consume_probe_results` does, and it always overwrites with `source="probe"` on a freshly parsed env. So a probe source from the wire is technically forgery, but it's hard to construct a useful forgery because you can't fake the parsed env without also having the tool round-trip succeed. Leave probe alone.)

The test `test_metadata_source_config_is_relabeled_to_request` is the closest analogue — a parallel `test_metadata_source_host_is_re_promoted_or_ignored` (or `_rejected`) would prove the partial fix.

Fledgewing:
Reviewed PR #66 at the new HEAD (86d843f) and posted 3 should-fix findings — see inline comments.

The fixer addressed the round-1 issues at the surface: probe=required now raises and translates to HTTP 503 with Retry-After: 5 (handle_fresh + resume_gone_session); metadata spoofing is mitigated by a new `from_wire_metadata` that re-labels claim-as-config -> request; the probe race is closed by a module-level `PROBE_LOCK` plus a `_PROBE_PENDING` sentinel published before the lock is released; and `host_hint` is gone from `resolve()` and its call sites. Test counts match the round-2 expectations exactly: caller_env 38 (+7), mcp_bridge 47 (+3), cli_bridge 44 (+0). All new tests inserted before the `__main__` runners. No FAILs in the offline suite.

What the fix got right:
- `CallerEnvironmentRequired` exception raised only at the right place (after forced config / request / probe / host branches fall through), not in the success paths.
- Probe lock body is fully synchronous (no `await` inside `async with PROBE_LOCK:`), so no deadlock surface. The `_PROBE_PENDING` sentinel is published BEFORE the lock is released, which is the correct ordering for atomicity. The asyncio.gather(3x same-fingerprint) test pins the count distribution.
- Text path's `required` -> `auto` downgrade is a sensible text-path-only shim (`dataclasses.replace` produces a local copy, ce_cfg is untouched), with `except CallerEnvironmentRequired: env = None` retained as defence-in-depth.
- The relabeling of a legitimate gateway stamp (forced config produces source=config upstream; the sidecar receives source=config, relabels to source=request) does NOT cause spurious probe=required refusals because env.source != "unknown" still skips the probe path on handle_fresh.

What the fix missed:
1. **`probe: required` refusal uses 503 + Retry-After: 5, but the failure mode is not transient.** The existing 503 patterns in this file (lines 1080, 1530) are for capacity / resume-gone situations where retrying might succeed because a slot freed. probe=required unresolvable won't resolve in 5 seconds unless the operator just edited plans.yaml. 400 (Bad Request) would be more semantically accurate.
2. **`from_wire_metadata` is more permissive than the code it replaced.** Stamps with values but no source (or source="unknown"/"made_up") now return parsed and are used as-is by both bridges; the old code fell through to `parse_request(body)`. A future refactor that "fixes" this case to fall through has no test coverage to catch the change.
3. **The metadata-spoof fix is partial.** `source="config"` is re-labeled (correctly — only the operator's plans.yaml can produce that label), but `source="host"` is ALSO derived from plans.yaml (the `fallback_platform` branch at line 642), and `from_wire_metadata` honors it verbatim. Same exploit class as the round-1 config-spoof, just at a different tier.

No regressions in the no-op paths:
- `host_hint` removal is clean: no remaining references outside the test docstring.
- The lock doesn't serialize anything that wasn't already serialized by the dicts (FAILED_PROBES / RESOLVED_PROBES eviction).
- The hook stamp at `switchyard/hooks.py` is unchanged; it still catches the `CallerEnvironmentRequired` exception silently, and the sidecar's own resolve() call catches it and returns 503 — so the gate-level refusal still works end-to-end.

Verdict: comment-only review. Two of the four round-1 fixes are solid (probe race, host_hint). The other two land but with subtleties worth addressing before merge.

Fledgewing:
Reviewed PR #66 at the new HEAD (5484285) — all three round-2 findings are addressed cleanly. No new findings.

Verification of the three round-2 fixes:

1. **`probe=required` refusal uses 400 (no Retry-After).** `mcp_bridge/server.py` lines 1285 and 1518 both raise `HTTPException(status_code=400, detail={"error": {"type": "caller_environment_required", ...}})` with no `Retry-After` header. The comment on line 1278 explicitly notes that 503+Retry-After would actively mislead the caller because nothing about the request/caller/operator config can change in 5 seconds. Test `test_probe_required_unresolvable_returns_400_no_unknown_wording` (renamed from the round-2 _returns_503_ variant) pins 400 AND asserts `not "retry-after" in headers`.

2. **`from_wire_metadata` now relabels `source=host` to `source=request`.** `switchyard/caller_env.py` line 184: `if raw_source == "config" or raw_source == "host": source = "request"`. The docstring was updated to mention `cfg.fallback_platform` as the only path that legitimately produces source=host. Two new tests pin the behavior at both layers: `test_from_wire_metadata_relabels_host_to_request` (unit) and `test_metadata_source_host_is_relabeled_to_request` (end-to-end through mcp_bridge).

3. **`from_wire_metadata` is strict again — stamps with no source / "unknown" / unknown labels return None.** `switchyard/caller_env.py` lines 190-191: `else: return None`. The round-2 test was updated (`test_from_wire_metadata_rejects_malformed_input`) to assert None for no-source, source="unknown", and unknown labels. End-to-end test `test_metadata_stamp_without_source_falls_through_to_passive_parse` pins the critical case: a wire stamp with values but no source + a passive-parseable body — the passive parse wins (windows / `C:\Users\demo\p`), the stamp's `linux / /app/mcp_bridge` does NOT win. This is exactly the bug the change exists to prevent, verified end-to-end.

Final regression sweep across the full PR diff:

- **Precedence chain** (`config > request > probe > host > unknown`): intact in `caller_env.resolve()` lines 649-667; covered by 9 `test_resolve_*` tests.
- **Probe safety**: `PROBE_COMMAND` is still exactly `printf 'cwd=%s\nplatform=%s\nshell=%s\n' "$(pwd)" "$(uname -s)" "$SHELL"` — no env/printenv, only `$SHELL` enumerated. `_maybe_probe` does NOT acquire a gate slot; the synthetic response is a small JSON envelope that travels back to the caller for local execution.
- **PROBE_LOCK**: body is fully synchronous (`dict` lookups, `find_command_tool`, `mint_probe_call_id`, `remember_probe_env`) — no `await` inside the lock, so no deadlock surface. `_PROBE_PENDING` sentinel is published INSIDE the lock before release; the response is built OUTSIDE the lock where nothing touches shared state.
- **No-tools passthrough**: `cli_bridge._handle_chat` line 1524 still raises 400 with `type=tools_unsupported` when `body.get("tools")` is truthy.
- **429/413/502 contracts**: `429` for sidecar-at-capacity, `413` for E2BIG, `502` for spawn failures — all intact.
- **SYSTEM_MODE semantics**: env block is appended to the system string BEFORE profile flags so SYSTEM_MODE=replace semantics are untouched (cli_bridge `_handle_chat` line 1611 + the `test_text_path_system_mode_replace_argv_unchanged_apart_from_system` test).
- **docker-compose.yml**: untouched (`git diff main...5484285 -- docker-compose.yml` produces empty output).
- **Tests before __main__ runners**: 40 / 49 / 44 tests properly registered; the `__main__` runners at lines 700 / 2444 / 1175 collect them via `sorted(globals().items())`.
- **No real-provider calls**: `tests/conftest.py` blocks all non-loopback sockets; only 127.0.0.1 (stub HTTP servers, subprocess bridges) is allowed.

Test counts match the round-3 expectations exactly: caller_env 40 (+2), mcp_bridge 49 (+2), cli_bridge 44 (+0). `render_preview.py` still renders (10 plans, 19 models, 5 lanes, 18 quota windows, 4 probes). No FAILs in the offline suite.

PR #67, opened 2026-09-22
Fix /v1/messages success accounting

### LINKED ISSUE #51: /v1/messages fires no success logging event: the slot is never released and the tokens are never booked

Fledgewing:
Fixes #51.

### The bug

LiteLLM 1.101.0 only fires `async_log_success_event` / `async_log_failure_event` for the four classic chat/text call types (`completion`, `acompletion`, `text_completion`, `atext_completion`). For `/v1/messages` (Anthropic Messages API) LiteLLM routes the success through `async_post_call_success_hook` (buffered) or `async_post_call_streaming_iterator_hook` (streamed) instead, with the normal success logging event never firing on that path. So a request on `/v1/messages` reached Switchyard's pre-call hook but had **no** place to finish it: the slot stayed claimed until the staleness sweep and the tokens spent never reached the ledger.

Measured upstream by the reporter against a Claude Code tool loop: a 2-slot plan filled its seat, then refused its own follow-up call, with no real work in flight on that seat. Streaming is worse — a streamed reply also bypasses `async_post_call_success_hook`.

### The fix

**Allow-list (not deny-list):** every call type LiteLLM routes through *normal* success logging stays there; everything else — including `/v1/messages`, both buffered and streamed — finishes by hand from the post-call hook. A future LiteLLM call type that does not log fails *safe* (the post-call side finishes it) instead of leaking a slot. `call_type` is stamped on the routing context at claim time so the post-call side can tell which branch owns the release.

**Buffered /v1/messages:** `async_post_call_success_hook` calls `_finish_success` after stamping the served-deployment tag back onto the response body (existing behaviour). `_finish_success` is the new idempotent core: it stops the heartbeat, releases the slot, books the usage onto the plan ledger (Anthropic `input_tokens`/`output_tokens` *and* OpenAI `prompt_tokens`/`completion_tokens` accepted; `cache_read_input_tokens` + `cache_creation_input_tokens` fold into `prompt_tokens` so the OpenAI-shaped ledger sees the full input cost), notes throughput on the pacer, resets the served plan's transient-failure streak, and absorbs the provider's `x-ratelimit-*-remaining` headers. Same core for `_finish_failure` on the failure side.

**Streamed /v1/messages:** `async_post_call_streaming_iterator_hook` wraps the upstream iterator in `_stream_chunks`, which:
* parses every Anthropic SSE frame for usage (`message_start` carries the input/cache totals; successive `message_delta` frames carry the running output/cache totals — a later `message_delta` overwrites because it is the latest authoritative reading, not an addition, so a fragmented stream cannot inflate the total),
* tolerates bytes / str / dict / pydantic-shaped chunks (synthetic stream, agentic loop),
* never breaks the client stream — a malformed frame is logged and skipped.

The finalize (slot release + ledger + throughput + transient-failure reset + limit headers) is a **detached `asyncio.create_task`**, scheduled from both `try/except/else` arms of the iterator. Awaiting the finish inside the hook would block the FastAPI/Starlette `StreamingResponse` after the last `yield` and would be killed by `GeneratorExit` on a client hangup; the detached task survives both, so client cancellation mid-stream, upstream errors, and clean completion all release the slot exactly once. Reasoning-tag splitting still runs in the streaming hook (LiteLLM gives us only one place on the streamed path).

**Exactly-once protection:** on the rare LiteLLM build where `async_log_success_event` and `async_post_call_success_hook` both fire for the *same* request (litellm router-level retry / fallback can produce this on the buffered OpenAI routes), the second arrival sees the `ctx[_TERMINATED]` marker set by the first and short-circuits. The same marker guards `_finish_failure` for double-fire failures. Without the marker, the slot would be released twice and the ledger would book the request twice — exactly the double-accounting the task calls out.

### Why the symmetric gate for logged call types stays in `async_log_success_event`

LiteLLM *normally* fires `async_log_success_event` for `acompletion`, but a future LiteLLM version could start firing it for an `anthropic_messages` build too. The call_type gate at the top of `async_log_success_event` / `async_log_failure_event` keeps the post-call side as the single owner for those routes even on a future LiteLLM that fires both — no matter which path arrives first, the second is a no-op on the marker.

### Verification

* `for t in tests/test_*.py; do python3 "$t" >/dev/null || exit 1; done` — all suites pass.
* `python3 tests/render_preview.py` — passes (10 plans, 19 models, 6 lanes, 18 quota windows, 4 probes; portal HTML rendered).
* `tests/test_messages_hooks.py` — 20 cases covering: pre-call hook stamps `call_type`; buffered `/v1/messages` releases slot + books Anthropic-shaped and OpenAI-shaped usage; cache tokens fold into `prompt_tokens`; `response_cost` booked only on metered plans; streamed Anthropic usage extracted across `message_start` + `message_delta` (with the last-delta-wins property); chunk passthrough tolerates malformed SSE / comments; dict + pydantic chunk shapes; client disconnect mid-stream releases slot via detached task; upstream exception releases slot via detached task; `acompletion` not double-finished by the post-call hook; streamed `acompletion` slot stays claimed (owned by `async_log_success_event`); failure double-call short-circuits on the marker; `/v1/messages` failure releases via `post_call_failure_hook`; defensive: stray `async_log_success_event` on `anthropic_messages` is a no-op; `UNLOGGED_CALL_TYPES` constant contains the right entries; low-level SSE parser tolerates partial / empty / junk frames; picker returns to cap-1 after a messages finish; MiniMax-shaped HTTP 200 with `base_resp.status_code` failure is recorded as a failure (no usage booked); end-to-end pick → finish → pick-again.

### Deployment

No deployment performed. This branch is a worktree; the rebuild + redeploy from the main checkout (`scripts/apply.sh --build`) is up to the operator after merge.


switchyard/hooks.py
```diff
@@ -518,15 +643,179 @@ async def async_post_call_streaming_iterator_hook(
...
+        try:
+            async for chunk in response:
+                try:
+                    self._absorb_chunk_usage(chunk, collected)
+                except Exception:
+                    log.debug("stream usage parse skipped for one chunk",
+                              exc_info=True)
+                yield chunk
+        except BaseException:
+            self._schedule_stream_finalize(ctx, collected, request_data)
```
Fledgewing: **[blocker] Streamed error path can lose the failure record due to a race with the proxy's post_call_failure_hook**

`_stream_chunks` schedules the same detached `_finish_success` task from both the `try/except BaseException` and `else` arms (hooks.py:688-692 + `_schedule_stream_finalize` at 694). The PR description acknowledges that both arms must finalize (the streamed-error path needs the partial-usage bookkeeping) but does not address what happens when the proxy's `async_streaming_data_generator` (common_request_processing.py:3687-3696) ALSO fires `await proxy_logging_obj.post_call_failure_hook(...)` for the same exception — which is the documented LiteLLM behavior for upstream errors mid-stream.

The `_TERMINATED` marker is supposed to prevent double work, but it picks ONE winner based on order, and which one wins is not deterministic:

- LiteLLM's `post_call_failure_hook` (utils.py:2525) awaits `self.update_request_status(...)` BEFORE reaching the `_callback.async_post_call_failure_hook(...)` call at utils.py:2585. `update_request_status` itself awaits `internal_usage_cache.async_set_cache(...)` whenever `self.alerting is not None` (i.e. Slack alerting is configured — true on any production deployment that follows the documented Slack-alerting recipe).
- That first await yields control. The detached task scheduled at hooks.py:725 (`asyncio.create_task(_run())` was added to the ready queue in hooks.py:689/692, before the exception propagated up to `async_streaming_data_generator`) runs FIRST. `_finish_success` sets `ctx[_TERMINATED] = "success"` at hooks.py:475 BEFORE its own first await at hooks.py:479.
- When the proxy eventually gets to SwitchyardHandler's `async_post_call_failure_hook` (hooks.py:805) and calls `_finish_failure`, hooks.py:538 sees the marker already set to "success" and short-circuits.

Net effect for a streamed /v1/messages that errors AFTER at least one `message_start` or `message_delta` chunk:

1. `_finish_success` runs — usage is booked, transient-failure streak is **reset** (hooks.py:516), throughput is noted, limit headers are absorbed. All four are wrong on a failure.
2. `_finish_failure` is a no-op — no `ledger.record(plan, failed=True, …)` (hooks.py:549), no `_handle_failure` / verdict (hooks.py:837), no cooldowns applied.

I reproduced this against the PR's own wiring with FakeRedis and a single `await asyncio.sleep(0)` inserted between the streaming hook and the failure hook to model the `update_request_status` yield:

```
Final terminated marker: success
Has failures in ledger: False
```

With NO intermediate yield (alerting off), the marker comes out `"failure"` and the failure IS recorded — which is why the test suite, which doesn't go through LiteLLM's proxy and which never configures Slack alerting on the FakeRedis-backed `ProxyLogging`, passes cleanly. The bug is invisible from inside this repo's test harness and only surfaces when Slack alerting is configured on the live LiteLLM proxy.

A robust design either (a) does not call `_finish_success` from the error arm — let `post_call_failure_hook` carry the failure, and accept that partial usage on a hard mid-stream error is lost (it is bounded; the alternative below is worse); or (b) makes `_finish_success` and `_finish_failure` aware of each other so the success path on a streamed error does not reset the transient-failure streak and the failure path still books whatever usage was collected (pass `collected` from the streaming hook through `ctx` or a known key, and have `_finish_failure` consult it). The marker alone is not enough — the marker is a "first writer wins" race resolution, and the streamed-error path currently has two first-writers competing.

The test suite should also cover the streamed-error + post_call_failure_hook interaction (either by directly chaining them with an intervening yield, or by mocking the proxy's `post_call_failure_hook` to await before invoking the callback), so this regression cannot ship silently again.

Fledgewing:
Reviewed 1 finding: 1 blocker — see inline comment at https://github.com/Fledgewing/SwitchYard/pull/67#discussion_r4075326816 (commit c7d59bd). 

TL;DR: the streamed-error path has a race between the detached `_finish_success` task scheduled from `_stream_chunks`'s `except BaseException` arm and the proxy's own `post_call_failure_hook`. The `_TERMINATED` marker resolves the race as "first writer wins", and with any intermediate await between scheduling and `_finish_failure` (e.g., the `update_request_status` await that fires whenever Slack alerting is configured on the proxy), the detached task wins, sets the marker to `"success"`, books usage AND resets the transient-failure streak, then `_finish_failure` short-circuits and the failure is lost — no ledger failure row, no cooldown. The test suite passes because nothing in this repo's harness inserts that intermediate yield; I reproduced the race locally by inserting a single `await asyncio.sleep(0)` between the streaming hook and the failure hook.

PR #71, opened 2026-09-22
Fix token counting on Anthropic tool schemas with items-less arrays (#59)

### LINKED ISSUE #59: Token counting fails on every Anthropic tool schema: 462 errors in 12 hours
Two lines that fire on essentially every tool-carrying request. Neither changes an answer, but between them they are most of what the gateway log says, which is its own problem - the `UnboundLocalError` from #48 was sitting in the same log five times and I only found it by grepping for it.

### 1. `failed to count tokens. Got - 'items'` - 462 times in 12 hours

```
LiteLLM Router:ERROR: router.py:12210 - litellm.router.py::_acount_pre_call_check_tokens:
failed to count tokens. Got - 'items'
```

It is a `KeyError` in LiteLLM's own schema walker, `litellm/litellm_core_utils/token_counter.py`:

```python
elif type == "array":
    # items is required, OpenAI throws an error if it's missing
    return f"{_format_type(props['items'], indent)}[]"
```

The comment is true of OpenAI and not of Anthropic: the Anthropic tool spec does not require `items` on an array, and Claude Code sends array parameters without it routinely. So every request carrying such a tool raises before a count exists, and the router's pre-call context-window check is skipped.

For SwitchYard specifically I think the consequence is only the log line - the pre-call hook has already rewritten the lane to one concrete deployment, so the router is not choosing anything and has nothing to filter. Worth saying out loud in case that is not the intent: if any part of the design expects that check to fire, it has never fired for a Claude-protocol client.

Two ways out, and I did not want to pick one for you:

* report it upstream (it looks like a clean fix there: `props.get("items") or {}`), or
* normalise the schema inside SwitchYard before LiteLLM sees it, which means touching the tool definitions on their way to the model, which I would not do without your say-so.

### 2. `CountTokens handler: 404 for /v1/responses/input_tokens` - 12 times in 12 hours

```
LiteLLM:ERROR: HTTP error in CountTokens handler: Client error '404 Not Found'
  for url 'http://claude-max-sidecar:8081/v1/responses/input_tokens'
LiteLLM Proxy:WARNING: Provider token counting failed (404). Falling back to local tokenizer.
INFO: "POST /v1/messages/count_tokens?beta=true HTTP/1.1" 200 OK
```

LiteLLM asks the deployment to count the tokens, the bridge has no such route, it falls back and answers 200. Working as designed, and the fallback is the honest behaviour since the bridges cannot count without spending a turn.

The one place it is not purely cosmetic: a client that calls `/v1/messages/count_tokens` to decide when to compact its own context is getting a local estimate rather than the vendor's number. Small, but it is a number a client acts on.

If a route is wanted there, `/v1/responses/input_tokens` returning the same local estimate would at least be silent. Happy to send that if you want it; leaving it alone is also a defensible answer, in which case the 404 is expected and could be logged as such.

Fledgewing:
Fixes #59.

## What was happening

The 462x/12h `failed to count tokens. Got - 'items'` lines in the gateway log come from `litellm.litellm_core_utils.token_counter._format_type`. Its `type == "array"` branch does a bare-subscript `props['items']`, which raises `KeyError: 'items'` on every Anthropic-format tool whose array property omits `items`. The Anthropic tool spec does not require `items` on an array property — `{"type": "array"}` is a complete schema — but litellm's counter was written against the OpenAI shape, which does. So every `tools=[...]` request from a Claude-protocol client took `token_counter` down on first use, and the gateway's pre-call context-window check never fired on tool-bearing traffic at all (it had silently never been reaching the count check in the first place). Upstream is not fixed as of 2026-09-22 (checked litellm `main`, commit `0abd9267c1`), so a pin bump would fix nothing and would violate the PIN POLICY at the top of `Dockerfile.gateway`.

## The fix

Three layers, each guarded against regression:

1. **Build-time patch in `Dockerfile.gateway`** on the digest-pinned litellm v1.101.0: replaces the bare-subscript with a tolerant `props.get('items') or {'type': 'string'}`, so an items-less array counts as `string[]` instead of raising. The exact-occurrence assert means a future pin bump that drifts past the patched line fails the build loudly rather than bringing the same bug back at request time with no commit message.
2. **Startup audit in `switchyard/selfcheck.py`** (`_audit_token_counter_patch`, wired into `_static_audit`): checks the vulnerable line is gone from the installed litellm, that `{"type":"array"}` now renders `string[]`, and that the well-formed `items=string` path is unchanged.
3. **Offline regression test in `tests/test_litellm_patch.py`** (4 tests, no `litellm` import): Dockerfile block intact, patched `_format_type` tolerates items-less arrays, unpatched `_format_type` still raises, and `selfcheck.py` wires the audit into `_static_audit()`. The test pins both the vulnerable fragment and the replacement, so a Dockerfile drift fails the offline suite, not the running image.

## Side effect worth calling out

The router's pre-call context-window check and the documented `context_window_fallbacks` had silently never fired for tool-bearing Claude-protocol traffic — the count was always raised away before the gate ran. With the patch they work as documented. Missing-`items` arrays are counted as string items, which is a small estimate delta but acceptable for a guard check.

## On the `CountTokens handler: 404 for /v1/responses/input_tokens` warning

Deliberately left as-is. LiteLLM's CountTokens handler posts there to ask a vendor for an exact token figure; the bridge has no such route (only `/usage`, `/health`, `/v1/models`, `/v1/chat/completions`), so the call 404s. LiteLLM falls back to its local tokenizer and answers `/v1/messages/count_tokens` with 200, which is what the patch enables. The bridges cannot count vendor tokens without spending a turn, so the local fallback is the honest number rather than a wrong one — no sidecar change is planned. Documented as expected in README.md and TESTING.md.

## Post-deploy verification

Manual step in `TESTING.md`. **Rebuild/redeploy from the main checkout only — never from a worktree** (`docker-compose.yml` pins `name: switchyard`, so a worktree `docker compose build` rebuilds and recreates the live containers):

```bash
# from the main checkout
scripts/apply.sh --build

docker compose exec gateway python3 -m switchyard.selfcheck
```

Expect all five `ok` lines, including:

```
switchyard.selfcheck:   ok  litellm._format_type tolerates items-less arrays; items=string still renders 'string[]'
```

Then exercise the patched path on a live request — full curl examples in TESTING.md §8.

## Files touched

- `Dockerfile.gateway` — build-time litellm patch block with PIN POLICY guard
- `switchyard/selfcheck.py` — `_audit_token_counter_patch` + `_static_audit` wiring
- `tests/test_litellm_patch.py` — new offline regression (4 tests)
- `README.md` — troubleshooting entry for the `failed to count tokens` log line
- `TESTING.md` — post-deploy manual verification + expected-404 note


README.md
```diff
@@ -629,6 +629,15 @@ Where a CLI has no system-prompt flag at all (OpenCode, Codex as configured),
 the caller's system text is folded into the top of the prompt rather than
 discarded.
 
+### `failed to count tokens. Got - 'items'` in the logs
```
Fledgewing: **[nit] Troubleshooting entry wedged between unrelated sections**

This new "### `failed to count tokens. Got - 'items'` in the logs" entry is placed between the end of the prompt-stacking discussion (CLI prompt handling ends at line 630) and "## OAuth subscriptions" (line 641). The two neighbouring topics have nothing to do with token counting, so a reader skimming for the symptom lands in the middle of an OAuth section without a heading bridge.

The natural home is right next to "### Litellm pin and self-check" (line 569), which already documents the four pre-existing selfcheck audits and the PIN POLICY that this entry also invokes by reference. Either append this entry to that section (rename the existing one to "Litellm pin, self-check and the token-counter patch", then add a sub-subsection), or move it just below "### Litellm pin and self-check" so the reader who follows the litellm-pin cross-reference lands in the same paragraph instead of jumping to a top-level OAuth heading.

Content is correct; just placement.

TESTING.md
```diff
@@ -841,3 +929,12 @@ them first if something misbehaves:
 7. **Real token allowances.** Everything works without them, but pacing stays
    idle and headroom stays estimated until either you set them or a plan hits a
    wall once and the observed-allowance learning records it.
+8. **`CountTokens handler: 404 for /v1/responses/input_tokens` from the
+   claude-max-sidecar is EXPECTED.** LiteLLM's CountTokens handler posts there
+   to ask a vendor for an exact token figure; the bridge has no such route
+   (only `/usage`, `/health`, `/v1/models`, `/v1/chat/completions`), so the
+   call 404s. LiteLLM falls back to its local tokenizer and answers
+   `/v1/messages/count_tokens` with 200, which is what step 8 above
```
Fledgewing: **[nit] "step 8 above" is now ambiguous — two sections both numbered 8**

After this PR the top-level TESTING.md has TWO sections that contain a "8":

- `## 8. Post-deploy check: the token-counter patch` (line 784)
- item `8. CountTokens handler: 404...` under `## Not verified here, and what to watch` (line 932)

This line says "which is what step 8 above verifies", which now reads as referring to itself — the only "8 above" inside the "Not verified here" list is the very bullet that contains this sentence. A reader following the breadcrumb lands back where they started.

Rephr to the section title rather than the position number, e.g.:

> ... which is what the "Post-deploy check: the token-counter patch" section above verifies.

or:

> ... which is what the §8 verification above proves.

so the link survives a future renumbering.

switchyard/selfcheck.py
```diff
@@ -305,6 +314,115 @@ def _callee_names(node):
...
+            f"verify the build-time patch: {type(exc).__name__}: {exc}"
+        )
+        return
+
+    pkg_dir = os.path.dirname(litellm.__file__)
+    src_path = os.path.join(
+        pkg_dir, "litellm_core_utils", "token_counter.py"
+    )
+    try:
+        with open(src_path, encoding="utf-8") as fh:
```
Fledgewing: **[nit] `open()` here lacks `errors="replace"` the sibling audit uses**

The neighbouring `_audit_proxy_pre_call_hook_present` opens every `.py` under `litellm/proxy/` with:

```python
with open(path, encoding="utf-8", errors="replace") as fh:
```

(line 282). This new audit opens `token_counter.py` with just `encoding="utf-8"` — line 353. In practice the file is litellm-bundled UTF-8 so this never bites today, but the asymmetry is one of those things future-you notices when the audit crashes on a UTF-8/UTF-16 mix in a pin-bumped upstream: the `UnicodeDecodeError` propagates out of the `try: open(...)` block (it is not an `OSError`) and the audit dies with a traceback instead of a clean `CRITICAL` line, taking the whole `_static_audit()` with it.

Adding `errors="replace"` matches the neighbour and keeps the failure mode consistent with every other audit in this file. One-character change.

Fledgewing:
Reviewed PR #71 at 39c672e: 3 nits, 0 blockers — see inline comments.

Layered fix (build-time patch + startup audit + offline test) is internally consistent and the test pins every layer. Patch logic verified by simulation: items-less, items=null, items={}, items=string, items=integer, items=boolean all render correctly with the patched line; items=integer renders `number[]` because the pinned litellm collapses integer+number into one return — the test acknowledges this asymmetry in a comment, which is the right call.

Findings (all nit severity, none blocking):
- README: README troubleshooting entry placement
- TESTING: "step 8 above" cross-reference is now ambiguous
- selfcheck: open() missing errors="replace" the sibling audit uses

Fledgewing:
Re-reviewed at a99ec28. All three round-1 nits are properly addressed:

- README: troubleshooting entry moved to line 614, immediately after "Litellm pin and self-check" — the natural home I asked for. Section content unchanged.
- TESTING: cross-reference rewritten to "the \"Post-deploy check: the token-counter patch\" section above" — unambiguous, survives any future renumbering.
- selfcheck: `_audit_token_counter_patch`'s `open()` now uses `errors="replace"`, matching the sibling `_audit_proxy_pre_call_hook_present`.

No new findings introduced by a99ec28: the diff between 39c672e and a99ec28 is exactly those three surgical fixes (+18/-9 README, +5/-4 TESTING, +1/-1 selfcheck). tests/test_litellm_patch.py still passes 4/4 in this worktree; full offline suite green except the two pre-existing failures (test_routing.py, test_verdict.py) that depend on a litellm install the worktree does not have.

All three previous threads resolved. Sign-off at this HEAD.

PR #72, opened 2026-09-22
gateway: per-CLI tool blocklist + first-turn proxied-session system note (fixes #63)

### LINKED ISSUE #63: Resumed CLI sessions keep reasoning in local tools, then fall off a cliff the moment they try to use them
Resumed CLI sessions reason as if their native tools still work, then fall off a cliff the moment they try to use them

## What happens

OpenCode session running through SwitchYard. Operator restarts the server while a session is mid-flight; the gateway picks the session back up on the next request (same `x-switchyard-session` header → same fingerprint → same lease after the mcp-bridge rebuild in `sidecars/mcp_bridge/server.py:1136-1199`). The model answers "hi" normally and immediately tries to run:

```
python3 ~/.config/opencode/scripts/orchestrate.py settle --run run_… --dispatch ctx_… --timeout-ms 900000
```

That script is on the operator's machine; it is not mounted inside the sidecar container, so the command fails. The model then thinks the whole relay is broken and goes into a confused loop:

> "Blocked by an environment handoff mismatch: the active workspace has no `/home/oranode/.config/opencode/scripts/orchestrate.py` or Orca CLI, so I cannot safely monitor or continue the already-dispatched plan worker."

Operator has to manually inject the right behaviour ("You are proxied, and must use MCP calls to interact with tools") before the session recovers. Once told, the model uses MCP and the run completes normally.

Same shape on every CLI-backed plan, not just opencode: any resumed CLI session whose prior turns narrated *local* tool use (scripts the user ran, paths on the user's machine, shell commands) will keep reasoning in that mode after the restart, because the bridge replay (`mcp_bridge/server.py:1076-1124` `flatten_with_tool_history`) hands the model back its own narration as if those tools still work.

## Why the current fix is silent

The harness already strips every native tool:

- `sidecars/cli_bridge/harness/opencode.json:7-12` disables `bash`, `edit`, `write`, `read`, `grep`, `glob`, `list`, `patch`, `todowrite`, `todoread`, `webfetch`, `websearch`, `task`, `multiedit` for the `switchyard` agent.
- `sidecars/cli_bridge/server.py:301-311` passes `--disallowed-tools Bash,Edit,Write,Read,Glob,Grep,WebFetch,WebSearch,NotebookEdit` to Claude.
- `sidecars/mcp_bridge/server.py:497-539` rewrites the opencode disable list per-request.

So the *capability* loss is real. What is missing is **any signal back to the model that the loss happened**. After a server restart, the rebuilt session loads a fresh `opencode.json`, every native tool is gone, but the model's *memory* is full of "I ran this command" turns. Nothing in the conversation prefix says "your tool surface has changed; use the MCP tools exposed under `mcp__switchyard__…` / `switchyard_…`". The first generation is therefore almost always wrong, and on a resumed session the wrong generation usually means a broken run.

The picker already filters at the gateway by capability, not by tool name (`switchyard/picker.py:178-179` checks `Plan.can_use_tools`, which is just `supports_tools: false`). There is no per-tool filter at any layer — `hooks.py:255` only sets `needs_tools = bool(data.get("tools"))` and forwards the list as-is.

The gateway also has no way to know which CLI is on the other end. `session.py:40-58` derives session identity from headers/metadata/fingerprint, but CLI identity is never read — every CLI looks like "an OpenAI-shaped request from a master key".

## Suggested fix

Two changes, both in `switchyard/hooks.py`. Both land before the picker (so they apply to every plan, every bridge) and are hot-reloadable for policy; only the new `Settings` field needs a rebuild.

#### 1. Per-CLI tool blocklist

Add a `cli_tool_block: dict[str, tuple[str, ...]]` field to `Settings` (`switchyard/models.py:153-185`) parsed in `load()` (`models.py:713-758`). Shape:

```yaml
settings:
  cli_tool_block:
    opencode:    [bash, edit, write, read, grep, glob, list, patch, todowrite, todoread, webfetch, websearch, task, multiedit]
    claude-code: [Bash, Edit, Write, Read, Glob, Grep, WebFetch, WebSearch, NotebookEdit]
    codex:       []
```

The hook identifies the calling CLI by reading a new header `x-switchyard-cli` (with `metadata.switchyard_cli` as fallback), the same way `session.py:43-51` reads session identity. Add it to a new `CLI_HEADERS` constant in `session.py` or a local helper in `hooks.py`. The header is set by the operator's CLI launcher, the same governance as `x-switchyard-session`.

In `async_pre_call_hook` (`hooks.py:251` area, right after `derive_session`), filter `data["tools"]`:

```python
cli = _identify_cli(data)
if cli and (blocked := self.registry.settings.cli_tool_block.get(cli)):
    before = data.get("tools") or []
    after  = [t for t in before if _tool_name(t).lower() not in {b.lower() for b in blocked}]
    if len(after) != len(before):
        log.info("cli=%s dropped %d tool(s): %s", cli, len(before) - len(after),
                 ",".join(_tool_name(t) for t in before if _tool_name(t).lower() in {b.lower() for b in blocked}))
    data["tools"] = after
```

`_tool_name` covers both OpenAI (`{"type":"function","function":{"name":…}}`) and Anthropic (`{"name":…}`) shapes. `needs_tools` is re-derived from `data["tools"]` so the picker does not route an emptied request to a tool-only lane. Drop is **loud** (single log line per dropped tool), never silent — silent drops are how we got here.

Why at the gateway rather than the sidecar: the gateway already owns policy (`Settings`), the picker, and the slot table; the sidecar's per-CLI disable lists are applied *inside* the CLI subprocess and can't tell the model anything changed. Doing it at the boundary also means a single tool the CLI shouldn't be calling cannot slip through on a future plan shape we haven't seen yet.

#### 2. First-turn system injection after pickup

Right before `return data` in `async_pre_call_hook` (`hooks.py:315`), prepend a system message on the first turn of a freshly picked-up session, and mark the session in Redis so it doesn't repeat:

```python
session_id = session
cli        = _identify_cli(data)
if cli in {"opencode", "claude-code", "codex", "gemini"} and session_id:
    if not await self._already_injected(session_id):
        note = {
            "role": "system",
            "content": (
                "You are running through SwitchYard, an LLM proxy. Your native "
                "tools (file edits, shell, etc.) are stripped and unavailable in "
                "this environment — they fail silently when called. Use the MCP "
                "tools exposed by this session to interact with the user's "
                "workspace; native CLI commands will not reach the user's machine."
            ),
        }
        msgs = list(data.get("messages") or [])
        i = 0
        while i < len(msgs) and msgs[i].get("role") == "system":
            i += 1
        msgs.insert(i, note)
        data["messages"] = msgs
        await self._mark_injected(session_id)
```

`_already_injected` / `_mark_injected` are tiny Redis helpers using a new key `K_INJECTED = "sy:inject:{session}"` set with the same TTL as `K_LEASE` (`lease_ttl_seconds = 1800` from `models.py:156`). Lease drop = injection-eligible again, which is the right behaviour: a fresh dispatch after a prior session was force-closed should also see the note.

The injection point has to be after the picker (we need `pick.sticky` to know whether we honoured an existing lease) and before `return data` (so `cli_bridge` sees it via `fold_system` at `cli_bridge/server.py:919-929` and `mcp_bridge` preserves it through `flatten_with_tool_history`). Both bridges already handle a leading system message correctly, so no sidecar code changes.

#### Tests go in `tests/`, before the runner block

Three tests for the injection (new `tests/test_session_injection.py`, following the `_build()` wiring pattern at `tests/test_verdict.py:42-71`):

1. first turn after pickup gets the system message;
2. second turn with the same session id does not re-inject (using `tests/fake_redis.py`);
3. a non-proxied CLI (openai-direct / grok) does not get the message;
4. no session id → no injection.

Two tests for the blocklist, in `tests/test_routing.py` next to the existing tool-capability tests at lines 320-416:

1. tool whose name is in the CLI's blocklist is dropped, the rest pass through;
2. empty `tools` after filter re-derives `needs_tools` so the picker does not misroute.

Reminder from CLAUDE.md: insert before the `if __name__ == "__main__":` runner — anything appended after is defined too late to be collected and the count silently does not go up.

## Why both, not just one

The blocklist alone stops the model from *calling* local tools, but it doesn't stop it from *thinking* in local tools — the narration in the resumed conversation still says "I ran the orchestrate script", and the first generation will keep trying to do the same thing until something tells it otherwise. The injection alone doesn't prevent the model from hallucinating tool calls that the bridge will then have to refuse. Together: the injection tells the model the new rule, and the blocklist is the backstop that catches any model that ignores it.

Happy to PR either piece; both are small and self-contained.

Fledgewing:
## Problem

Issue #63: after a server restart, a resumed CLI session's conversation history still narrates local tool use (scripts, user-machine paths), so the model keeps trying native tools that the bridge harness has stripped, fails, and loops confused until the operator manually explains. The capability loss was real but silent — no signal back to the model.

## Fix

Two halves, both at the gateway in `switchyard/hooks.py` **before the picker** runs (so every plan and bridge sees the cleared request, and only the new `Settings.cli_tool_block` field requires a rebuild after hot-reload of `config/plans.yaml`):

1. **Per-CLI tool blocklist.** `Settings.cli_tool_block` keyed by the lowercased `x-switchyard-cli` header (fallback `metadata.switchyard_cli`). Defaults mirror what each bridge harness already disables internally:
   - `opencode` — its script-aware tools
   - `claude-code` — its native run/edit/file tools
   - `codex` — empty (it doesn't carry native tools)

   Matching is case-insensitive over both OpenAI (`tools[].function.name`) and Anthropic (`tools[].name`) tool shapes. Drops are logged loudly in a single line naming every dropped tool. `needs_tools` is **re-derived after filtering** so an emptied request routes as plain text instead of 429ing a tool-only lane.

2. **First-turn system injection on session pickup.** A short note telling the model its native tools are stripped and it must use the session's MCP tools, inserted *after the leading system messages* on the first turn of a freshly picked-up session. Deduplicated by a Redis marker `sy:inject:{session}` whose TTL matches the lease; `drop_lease` clears the marker so a hard plan rejection re-enables injection on the next pick. Deliberately **NOT** gated on `pick.sticky`: in the production incident Redis died with the stack, so the resumed pick was non-sticky — and a sticky gate would have never fired, which is precisely the bug.

## Config

- `cli_tool_block` documented in `config/plans.example.yaml` (the tracked example).
- Operator live `config/plans.yaml` untouched.
- No `docker-compose.yml` changes.
- No sidecar changes needed (both bridges already preserve a leading system message).

## Tests

8 new, full offline suite green:

- `tests/test_session_injection.py` — **new file**, 6 tests:
  - opencode gets the system note on first turn, marker set
  - same-session second turn does NOT re-inject
  - non-proxied CLI (e.g. `openai-direct`) gets no message
  - message-less request: nothing touches
  - lease drop re-enables injection
  - metadata fallback (`switchyard_cli`) when header missing
- `tests/test_routing.py` — **+2 tests**, growing 63 → 65:
  - loud case-insensitive drops (OpenAI + Anthropic shapes)
  - `needs_tools` re-derived: an all-blocked `tools` array lets the request into a no-tools lane

## Deployment note

`switchyard/` is baked into the gateway / portal / sidecar images, not mounted. After merge, rebuild and redeploy happens **from the main checkout** (`scripts/apply.sh --build`), never from a worktree — worktrees address the same `name: switchyard` compose project, so building or recreating from a worktree rebuilds the live containers.

Closes #63.


switchyard/models.py
```diff
@@ -562,6 +597,31 @@ def _listify(mapping: Any) -> dict[str, list[str]]:
...
+        raise ValueError(
+            f"settings.cli_tool_block must be a mapping of cli -> tool list, "
+            f"got {type(raw).__name__}: {raw!r}")
+    out: dict[str, tuple[str, ...]] = {}
+    for cli, names in raw.items():
+        if not isinstance(names, list):
+            raise ValueError(
+                f"settings.cli_tool_block[{cli!r}] must be a list of tool "
+                f"name strings, got {type(names).__name__}: {names!r}")
+        out[str(cli)] = tuple(names)
```
Fledgewing: **[should-fix] `_parse_cli_tool_block` doesn't actually loud-parse — accepts non-string list entries and does not normalize CLI keys**

The docstring on lines 601-607 advertises "Loud-parse on purpose, matching the rest of `load()`" and explains the rationale: "a typo that drops the default to `{}` would silently let every CLI's native tools reach the wire". The implementation only enforces half of that — two silent-failure paths slip through `tuple(names)` on line 621:

1. **Non-string entries inside the inner list are silently accepted.** A config like `cli_tool_block: {opencode: [bash, 123, null, Edit]}` parses to `{opencode: ('bash', 123, None, 'Edit')}` and the gateway then crashes with `AttributeError: 'int' object has no attribute 'lower'` on the first matching request (in `{n.lower() for n in ...}` at hooks.py:289). The intent of the loud-parse is to surface the typo at startup, not at the first filtered request. Fix: also `raise ValueError` when any element of `names` is not a string, e.g. `if not all(isinstance(n, str) for n in names): raise ValueError(...)`.

2. **CLI keys are not normalized to lowercase.** `identify()` (session.py:92) returns the lowercased CLI name, and `plans.example.yaml` says "Keys are matched against the lowercased header value — keep config keys lowercase." Yet the parse stores `out[str(cli)] = tuple(names)` preserving whatever case the YAML supplied. So `Opencode: [bash, edit]` produces a dict keyed by `"Opencode"`; the hook then does `cli_tool_block.get("opencode", ())`, misses, and silently disables filtering for opencode — exactly the silent capability-loss the loud-parse is supposed to prevent. Fix: lowercase the key here, e.g. `out[str(cli).strip().lower()] = tuple(names)`; or `raise ValueError` on a non-lowercase key.

I confirmed both behaviors against the live parser: `_parse_cli_tool_block({'opencode': [123, None, 'valid']})` returns `{'opencode': (123, None, 'valid')}` without complaint, and lowering the dict key by hand produces the empty-filter behavior described above.

tests/test_session_injection.py
```diff
@@ -0,0 +1,374 @@
...
+
+
+if __name__ == "__main__":
+    n = 0
+    for name, fn in sorted(list(globals().items())):
+        if name.startswith("test_") and callable(fn):
+            print(f"{name}:")
+            fn()
+            n += 1
+    print(f"\n{n} injection tests passed")
```
Fledgewing: **[nit] Missing trailing newline at end of file**

Every other Python file in this repo (and every other test in this directory — `test_routing.py`, `test_verdict.py`, `test_classify.py`, …) ends with a newline. This file's last line is `print(f"\n{n} injection tests passed")` with no `\n` — the diff carries `\ No newline at end of file` for it. Trivial to fix and easy to drift on again; ask the editor to add one when you next touch the file.

Fledgewing:
Reviewed 2 findings: 0 blocker, 1 should-fix, 1 nit — see inline comments at 4505413.

Design-intent checks against issue #63 all hold in the diff: blocklist filter runs before the picker and `needs_tools` is re-derived after, drops emit a single loud log line naming every dropped tool, `data["tools"]` is only assigned inside the `if dropped` branch so an absent key never gets promoted, injection is gated on `cli ∈ INJECT_CLIS ∧ session ∧ ¬marker` and explicitly not on `pick.sticky`, the marker TTL reuses `lease_ttl_seconds`, `drop_lease` clears both the lease and the marker in a single Redis call covering all three drop sites, defaults are per-instance (`default_factory` builds a fresh dict of tuples), and `config/plans.example.yaml` documents the new key. Repo rules respected — no `.env`/`.env.example`/docker-compose/sidecar touched, operator-live `config/plans.yaml` is gitignored and untouched, new tests are defined before the `__main__` runner. Tests exercise what they claim: case-insensitive matching across both OpenAI and Anthropic shapes in `test_cli_blocklist_drops_blocked_tools_and_passes_the_rest`, the post-filter `needs_tools` re-derivation in `test_emptied_tools_re_derive_needs_tools_so_the_picker_does_not_misroute`, and the lease-drop re-enable path in `test_lease_drop_re_enables_injection`. Full offline suite green.

PR #73, opened 2026-09-22
opencode-go: preserve charged tokens on the dropped-text failure and stop cooldown escalation for it (#64)

### LINKED ISSUE #64: opencode-go/glm-5.3-flash drops the text event ~15-30% of the time; sidecar raises and tokens are silently lost from the ledger
## Symptom

OpenCode Go against `glm-5.3-flash` intermittently returns a `BadGatewayError` to the caller, with the sidecar's 502 detail attached:

```
litellm.BadGatewayError: BadGatewayError: OpenAIException - Error code: 502 -
{'detail': 'opencode cli yielded no parsed answer: no assistant text in CLI
output: line 1 column 1 (char 0)'}.
Received Model Group=sy.opencode-go.glm-5.3-flash
Available Model Group Fallbacks=None
```

Same error today on `sy.opencode-go2.glm-5.3-flash` (`switchyard-gateway-1` log, `17:53:02`, `18:02:32`). Same `BadGatewayError` shape on `sy.opencode-go.glm-5.3-flash` (`17:52:26`).

## Reproduction (direct CLI, no Switchyard in the path)

The OpenCode CLI itself is what fails: with `--format json`, it returns `step_start` and `step_finish` events but **no `text` event**, even though `step_finish` reports the model produced 17–27 output tokens (which the provider will bill). Two side-by-side 20-call sweeps:

```
opencode-go-sidecar,  opencode-go/glm-5.3-flash,  "Output a single character: K"  →  OK=14  BUG=6  (30%)
opencode-go2-sidecar, opencode-go/glm-5.3-flash,  "Output a single character: K"  →  OK=8   BUG=2  (20%)
```

A failed run looks like:

```
{"type":"step_start", "part":{... "type":"step-start"}}
{"type":"step_finish","part":{... "type":"step-finish",
  "tokens":{"total":2309,"input":239,"output":22,"reasoning":0,"cache":{"write":0,"read":2048}},
  "cost":0.00010829}}
```

The model spent 22 output tokens and the subscription was charged `$0.00010829`; the CLI reported neither a `text` event nor any other assistant-content event, so Switchyard cannot see what the model said.

The bug is specific to `glm-5.3-flash`:

```
opencode-go/glm-5.3      "Say OK"  →  OK=10  BUG=0
opencode-go/grok-4.6     "Say OK"  →  OK=10  BUG=0
opencode-go/glm-5.3-flash "Say OK" →  OK=17  BUG=3   (15%, with wide run-to-run variance)
```

OpenCode CLI is pinned to `opencode-ai@1.18.31` in `Dockerfile.sidecar:24`.

## What SwitchYard does on the failed path

1. `cli_bridge/server.py:853` raises `json.JSONDecodeError("no assistant text in CLI output", stdout, 0)` because `parse_output` saw no event with `part.type == "text"` (or any of the generic fallbacks).
2. `cli_bridge/server.py:1083-1086` catches it and returns 502 with `detail = "opencode cli yielded no parsed answer: ..."`.
4. `classify.py:244` classifies the 502 as `TRANSIENT` (`upstream 502`, 60s base cooldown), and the escalated ladder doubles it on each repeat (`60 → 120 → 240 …`).
5. The caller sees `BadGatewayError`. In this incident `opencode-go2` ran out the ladder (`streak=3 -> cooldown 240s`) before any request succeeded against it.

## Secondary bug: the bookkeeping is wrong

`parse_output` accumulates `usage` from `step_finish.tokens` while iterating, then **raises before returning it**:

```python
# cli_bridge/server.py:811-855
for evt in iter_json_objects(stdout):
    ...
    if tokens:
        add("output_tokens", tokens.get("output"))   # populates `usage`
        ...
if not text_parts:
    raise json.JSONDecodeError(...)                  # `usage` is discarded here
```

So the provider charged the subscription for those output tokens and Switchyard's `Ledger.record(failed=True, ...)` (hooks.py:401 / 450) records a **failure with zero tokens**. The subscription's dollars-per-turn denominator shrinks artificially; if the failure mode is reproducible across requests, the plan looks cheaper than it actually is. Worse: any caller retry (via `async_pre_call_hook`) re-enters the picker against an unchanged quota view.

## Why this matters

- **User-visible**: roughly 1 in 5–7 calls on `glm-5.3-flash` is a hard failure to the caller, not a transient retry that lands on a different plan. The `sy.*` model group has a single deployment, so even with `num_retries=1` the retry just hits the same broken path (same shape as #47, different cause).
- **Plan cooling escalates the outage**: a plan with a 30% drop rate escalates the cooldown ladder faster than the bug clears, and the plan sits parked for several minutes after the underlying CLI recovers.
- **Token accounting drifts**: a meaningful fraction of `glm-5.3-flash` tokens are charged by the provider but invisible to Switchyard's quota tracking. With the per-model economics dashboard (`effective_$/Mtok`) driven by those numbers, the plan can look like a bargain it isn't.

## Fix plan

In rough order of value:

1. **Return `usage` even on the no-text raise** (`cli_bridge/server.py:853`). Either surface a synthetic `{result: "", usage: <accumulated>, ...}` payload so the gateway's success hook books the tokens, or restructure the failure path so `_handle_failure` knows the charged token count and can `ledger.record(failed=True, prompt_tokens=…, completion_tokens=…)`. Closes the bookkeeping drift immediately.
2. **Detect "model answered but the CLI lost the text"** specifically: if `text_parts` is empty AND `usage["output_tokens"]` is > 0, the failure is **not** the CLI silently crashing — it's a model/CLI quirk that should not put the plan on the transient ladder. A targeted outcome (e.g. new `Outcome.MISSING_TEXT` with a short cooldown, NOT doubled) keeps the plan alive while we wait for upstream.
3. **Retry once before failing**. The parser already knows the model produced tokens; one same-process retry often produces a `text` event where the first attempt didn't. Avoid the round-trip back through the picker on the first failure.
4. **Track this in the parser, not the classifier**. The classifier cannot tell a 502 "sidecar at capacity" apart from a 502 "no parsed answer"; the sidecar knows and should label it before the gateway sees it (or the sidecar should include enough info in `detail` for the classifier to read).
5. **Pin or report**: confirm the upstream `opencode-ai` issue (an event with non-zero `output_tokens` and no `text` part should not be possible), pin to a fixed version when one ships, and add an offline test that drives `parse_output` with the bug-shaped stream and asserts the usage is preserved.

## Evidence appendix

- `docker logs switchyard-opencode-go-sidecar-1 | grep -c "502 Bad Gateway"` → 2 over 20 calls in the recent window; same for `opencode-go2-sidecar`.
- `docker logs switchyard-gateway-1 | grep -c "no parsed answer"` → 10 occurrences today, all on `glm-5.3-flash`.
- `docker logs switchyard-gateway-1 | grep "transient failure streak"` shows `opencode-go2` reaching `streak=3 -> cooldown 240s` on the back of three of these.
- CLI probe code used for the reproduction (run from the repo, not a deployed caller):
  ```bash
  for i in $(seq 1 20); do
    out=$(docker compose exec -T opencode-go-sidecar \
      opencode run --model opencode-go/glm-5.3-flash \
        --format json --agent switchyard \
        "Output a single character: K. Nothing else." 2>&1)
    echo "$out" | grep -q '"type":"text"' || echo "$out"
  done
  ```

Fledgewing:
## Problem

Issue #64: roughly 15-30% of `opencode-go/glm-5.3-flash` CLI runs emit a
`step_finish` event with non-zero output tokens but never emit a matching
`text` event, so the wrapper raises and the sidecar's structured parser
treats the stream as unparseable. Two consequences followed:

1. The 502 detail was the legacy "yielded no parsed answer" shape, which
   the gateway classified as a generic upstream 5xx → `TRANSIENT`. With
   `num_retries=0` at the router, every consecutive failure rode the
   doubling ladder (`bump_and_cool`: 60s → 120s → 240s → …), so a benign
   ~25% instantaneous failure mode parked the plan for minutes per streak.
2. The provider charged the request for those tokens, but the gateway's
   ledger recorded `failed=True` with zero usage. The effective $/Mtok
   denominator and the burn-rate timer both under-counted a chunk of
   real spend — invisible until the next reconciliation.

## Fix

Two workstreams land together so the contract stays in lockstep.

**WS1 — cli_bridge sidecar (`sidecars/cli_bridge/server.py`)**

- `parse_output` now raises `CliNoTextError(ValueError)` carrying the
  accumulated usage when `text_parts` is empty but charged tokens exist
  (non-zero `input_tokens`, `output_tokens`, or `provider_cost`).
- `_run_cli` splits the spawn-and-parse body into `_run_cli_attempt`,
  catches `CliNoTextError` exactly once, and retries the CLI **once** on
  that shape — no other failure mode triggers a retry (auth, quota,
  timeout, empty stdout, non-zero exit all stay single-attempt).
- A successful retry's payload has attempt-1's usage folded into it, so
  both attempts are booked against the plan exactly once each.
- A second no-text-with-usage failure raises the final 502 carrying the
  **combined** usage of both attempts in the literal contract shape:
  `{PROVIDER} cli emitted tokens with no text (prompt_tokens=N, completion_tokens=M)`.
- `CliNoTextError` is a `ValueError` so the structured-parser catch in
  `mcp_bridge` continues to route the underlying 502 path without any
  change there.
- A zero-usage no-text stream keeps the legacy `json.JSONDecodeError`
  path: no retry, original "yielded no parsed answer" 502 — gated on
  `extract_vendor_code` returning something for the contract counts
  being parsed at all.

**WS2 — gateway (`switchyard/classify.py`, `switchyard/hooks.py`)**

- `Outcome.TEXT_LOST = "text_lost"` — a new outcome distinct from
  `TRANSIENT`, so consecutive failures never feed the doubling ladder
  (`_apply_verdict`'s `verdict.outcome is Outcome.TRANSIENT` branch is
  the only path that calls `bump_and_cool`).
- Fixed 10-second cooldown, not `LONG`/defaulted: a sidecar quirk that
  self-heals on the next call does not need a minute-long sit-out.
- `_NO_TEXT` regex matches the WS1 contract phrase exactly, including
  the two parenthesised `prompt_tokens=`/`completion_tokens=` counts.
- `extract_no_text_tokens(message)` helper pulls the charged counts out
  using the **same** `_NO_TEXT` regex the classifier uses, so the two
  cannot drift apart.
- `hooks.async_log_failure_event` calls `extract_no_text_tokens` on the
  exception message; on a match it books `prompt_tokens` and
  `completion_tokens` into `ledger.record(failed=True, ...)`. Any other
  failure shape keeps the existing zero-token record.

## Test plan

- `tests/test_classify.py` (now 21 tests, was 17):
  - `test_opencode_text_lost_classifies_as_text_lost_with_ten_second_cooldown`
  - `test_text_lost_cooldown_does_not_double_across_consecutive_failures`
  - `test_ordinary_502_without_contract_phrase_still_classifies_transient`
    (regression guard: contract phrase outranks the generic 5xx rule)
  - `test_extract_no_text_tokens_returns_counts_for_contract_message`
  - `test_extract_no_text_tokens_returns_none_for_unrelated_prose`
- `tests/test_cli_bridge.py` (now 43 tests, was 39):
  - `test_no_text_with_charged_usage_preserves_usage_on_parse`
  - `test_final_502_detail_carries_combined_usage_in_contract_shape`
  - `test_retry_fires_only_on_no_text_with_usage_and_merges_first_usage`
    (auth/quota/empty stay single-attempt; successful retry folds
    attempt-1's usage into attempt-2's payload)
  - `test_zero_usage_no_text_does_not_retry_and_keeps_existing_message`
    (regression guard: zero-usage stays on the legacy path)

Offline suite (no provider, no Docker, no Redis):

```bash
for t in tests/test_*.py; do python3 "$t" >/dev/null || echo "FAIL $t"; done
```

Both `test_cli_bridge.py` and `test_classify.py` run silently green.

## Deploy notes

`switchyard/*.py` and `sidecars/*.py` are `COPY`ed into the gateway /
portal / sidecar images — not mounted. After this PR is merged, the
redeploy must run from the **main checkout** (not this worktree) with:

```bash
docker compose build gateway portal claude-max-sidecar xai-token-proxy
docker compose up -d
```

`docker compose restart` re-runs the OLD code and looks like the change
had no effect, so do not use it.

Fixes #64


switchyard/classify.py
```diff
@@ -230,6 +262,15 @@ def classify(
         return Verdict(Outcome.QUOTA_EXHAUSTED, secs, "quota exhausted (from message)", code)
     if _CONCURRENCY.search(msg):
         return Verdict(Outcome.CONCURRENCY, 20, "connection limit", code)
+    if _NO_TEXT.search(msg):
```
Fledgewing, line 265: **[blocker] This branch is only reachable from cli_bridge's emitted 502, not from mcp_bridge — opencode-go/glm-5.3-flash (the plan that exhibits issue #64) runs mcp_bridge, so the fix doesn't take effect in production.**

`docker-compose.yml` sets `BRIDGE: mcp` for both `opencode-go-sidecar` and `opencode-go2-sidecar` (lines 214, 241), so those sidecars run `mcp_bridge/server.py`, not `cli_bridge/server.py`. mcp_bridge loads `cli_bridge` as a module via importlib and calls `cli_bridge.parse_output(stdout, PROFILE["parser"])` directly (`mcp_bridge/server.py:734`) — it does NOT go through `cli_bridge._run_cli`.

When `parse_output` raises `CliNoTextError(usage)` (a `ValueError`), mcp_bridge catches it at `mcp_bridge/server.py:735` and emits the **legacy** 502 detail:

```python
detail=f"{PROVIDER} cli yielded no parsed answer: {exc}"
```

where `str(exc)` is `"no assistant text in CLI output (charged tokens present)"`. The full string the gateway sees is therefore:

```
opencode cli yielded no parsed answer: no assistant text in CLI output (charged tokens present)
```

That string does **not** contain the contract phrase `cli emitted tokens with no text (prompt_tokens=N, completion_tokens=M)`, so:

- `_NO_TEXT.search(msg)` (this branch) does not match.
- `extract_no_text_tokens(msg)` returns `None`.
- `classify(502, msg)` falls through to the generic 5xx rule → `Outcome.TRANSIENT, cooldown_seconds=60`.

Verified locally by running `classify(502, "<PROVIDER> cli yielded no parsed answer: no assistant text in CLI output (charged tokens present)")` against the post-PR classifier: it returns `Outcome.TRANSIENT, cooldown_seconds=60`, not `Outcome.TEXT_LOST, cooldown_seconds=10`. `extract_no_text_tokens` on the same string returns `None`.

Concrete consequences for issue #64 in production:

1. **No retry** — `_run_cli`'s retry lives in `cli_bridge`; mcp_bridge calls `parse_output` directly.
2. **No TEXT_LOST cooldown** — the plan rides the TRANSIENT doubling ladder (60 → 120 → 240 → …) on every dropped-text failure, the exact behavior issue #64 was meant to stop.
3. **No ledger booking** — `extract_no_text_tokens` returns `None`, so `hooks.async_log_failure_event` records `failed=True` with zero tokens (the secondary bug the issue called out).

The PR description acknowledges the design choice ("`CliNoTextError` is a `ValueError` so the structured-parser catch in `mcp_bridge` continues to route the underlying 502 path without any change there"), but misses that this decision breaks the contract-string-in-lockstep invariant for every plan running with `BRIDGE: mcp` — which today is all four sidecars in `docker-compose.yml`, including the two plans the issue is filed against.

Suggested fix: catch `CliNoTextError` explicitly in `mcp_bridge/server.py:735`, build the same contract phrase as `cli_bridge._run_cli` (move the f-string into a helper in `cli_bridge`, e.g. `format_no_text_detail(usage)`, and have both bridges call it), and emit it as the 502 detail so the gateway sees the same `Outcome.TEXT_LOST` shape it sees from `cli_bridge`.

Fledgewing:
Reviewed 1 inline finding (blocker) at commit 1fa0d85 — see inline comments on `switchyard/classify.py:265`.

One blocker: the WS1 contract phrase (`<PROVIDER> cli emitted tokens with no text (prompt_tokens=N, completion_tokens=M)`) is emitted only by `cli_bridge._run_cli`, but the opencode-go and opencode-go2 sidecars (the plans that exhibit issue #64) run `mcp_bridge/server.py` per `BRIDGE: mcp` in `docker-compose.yml`. mcp_bridge catches `CliNoTextError` as a plain `ValueError` and emits the legacy `"<PROVIDER> cli yielded no parsed answer: ..."` detail, which the classifier's `_NO_TEXT` regex does not match. The plan therefore rides the TRANSIENT doubling ladder on every dropped-text failure and the ledger does not book the charged tokens — the exact failure modes issue #64 was meant to fix.

Suggested fix: catch `CliNoTextError` explicitly in `mcp_bridge/server.py:735`, build the same contract phrase (factor out a `format_no_text_detail(usage)` helper in `cli_bridge`), and emit it as the 502 detail so the gateway sees the same `Outcome.TEXT_LOST` shape it sees from `cli_bridge`.

Offline suite: all 14 test files green. Test count claim: PR says `test_classify.py (now 21, was 17)`, actual `was 16` (off by one — cosmetic).

PR #79, opened 2026-09-22
picker+portal: rank perishable/lowest_utilization per provider family (fixes #53)

### LINKED ISSUE #53: perishable re-ranks across provider families: a ChatGPT overflow seat out-scored every Claude seat
Same reporter as #44. This is on top of the perishable strategy from #40, which I have been running.

## What happens

`_perishable_visit_order` re-ranks every member of the lane together. My lane is five Claude Max seats with one ChatGPT seat behind them as overflow. The ChatGPT member's score is computed on a different plan's scale, and it out-scored all five Claude seats, so the overflow seat got visited first while Claude capacity sat unused. A tail member of another provider is there to catch spill, not to compete.

## What I did here

Re-rank only the leading provider family, then append the rest in their configured order:

```python
fam = self.registry.plan_of(members[0]).provider_family if members else None
head = [m for m in members if self.registry.plan_of(m).provider_family == fam]
rest = [m for m in members if self.registry.plan_of(m).provider_family != fam]
members = _perishable_visit_order(head, by_ref, order, skipped) + rest
```

Eight lines, no behaviour change for a single-provider lane. Keeps the ordered fill that makes lane capacity the sum of the live plans' caps.

This may be better solved by normalising the score per plan, or by the nested strategies in #43 where the tail becomes its own stage. Either way, the current behaviour surprises anyone who mixes providers in one lane. Happy to PR the small version if you want it now.

Fledgewing:
## What

Provider-family partition for `perishable` / `lowest_utilization` re-ranking. The score no longer competes across `plan.provider_family`: refs are partitioned into one bucket per family, buckets ordered by first appearance in the lane body, scored refs in score-desc order within each bucket, unscored refs last. Applied at all three sites via a shared `family_partitioned_order` helper, with a tier offset baked into the stored `score_<ref>` values so the read-back sort (picker, board) reproduces the partition without re-applying it.

## Failure mode (issue #53)

A perishable lane mixing plans from different providers (e.g. `[claude-max/fable, openai/astra]`) interleaved on raw score. A high-room `openai/astra` briefly outranked every `claude-max/fable` candidate every time the writer fired, routing new sessions to a foreign plan at the moment the operator most expected the lane to keep the leading family in charge. The picker walked the score-sorted list verbatim; the per-group writer wrote the same interleaved hash; the board displayed the same cross-family rank. A quiet avalanche into a foreign plan, recurring on every probe poll.

## Design

- New helper `family_partitioned_order(refs, config_order, family_of)` in `switchyard/usage.py`:
  - Bucket order = first appearance of each family in `config_order` (the lane body, or the group's leaf refs in declaration order).
  - Within a bucket, the input order is preserved verbatim — the score-desc / unscored-last ordering the caller already produced carries through.
  - A `None` family joins one shared 'unspecified' bucket keyed as the empty string; the schema makes `provider_family` either a non-empty str or `None`, so the empty-string key never collides with a real family.
  - A single-family input is byte-identical to the input — the regression bar.

- Three call sites use it:
  - **portal lane writer** `_recompute_perishable_for_plan` in `switchyard/portal/app.py`
  - **portal per-group writer** `_write_group_order` in `switchyard/portal/groups.py`
  - **picker walk** `_visit_order_ranked` in `switchyard/picker.py`

- Tier offset (`_TIER_OFFSET = 1e9`) added to the stored `score_<ref>` values:
  - The reader sorts by score desc, so a per-family tier above the actual score range (room_pct ≤ 100) preserves the partition when the picker / board read the hash back without re-applying the helper.
  - Previous-poll entries have the offset stripped before reuse, so the offset doesn't compound across polls.
  - The picker re-partitions explicitly anyway — the tier offset is defence-in-depth, not the only partition mechanism.
  - `_one_adjacent_swap_partitioned` runs the swap **per family bucket**: a ref cannot move across a bucket boundary in either direction. A single-family set produces one bucket, so its result matches the legacy swap exactly.

- Deliberate behaviour change for mixed-family lanes only. Single-family lanes are bit-for-bit identical to today's output — the regression bar pinned by the two single-family tests.

- The capacity board's displayed ranking order (read via `get_group_order` / `get_lane_order`) now shows the family partition too, because the tier offset carries it through the score-desc sort the board performs.

## Test evidence

- `tests/test_routing.py` grew from 68 → 78 tests (+10 new regression tests, all green).
- `tests/test_portal.py` stayed at 17 tests; the two existing cross-family expectations were updated to pin the family-partition answer (the old assertion `["openai/astra", "claude-max/fable"]` was the bug).
- The 10 new tests, all in `tests/test_routing.py`:
  - `test_family_partitioned_order_single_family_is_identity` — regression bar for the helper itself.
  - `test_family_partitioned_order_separates_families_within_input_order` — multi-family partition preserves within-bucket input order.
  - `test_family_partitioned_order_interleaved_declaration_keeps_bucket_order` — declaration order A,B,A produces buckets A then B (not interleaved).
  - `test_family_partitioned_order_null_family_joins_unspecified_bucket` — `None` family buckets together, real families stay separate.
  - `test_perishable_writer_keeps_chatgpt_style_overflow_behind_claude_family` — the issue-#53 scenario at the lane writer.
  - `test_per_group_writer_perishable_mixed_family_ranks_leading_first` — same at the per-group writer, `perishable` strategy.
  - `test_per_group_writer_lowest_utilization_mixed_family_ranks_leading_first` — same shape, `lowest_utilization` strategy.
  - `test_picker_visit_order_ranked_keeps_leading_family_ahead_of_higher_other` — the picker's walk also partitions.
  - `test_perishable_lane_writer_single_family_is_byte_identical_to_pre_partition` — regression bar at the lane writer.
  - `test_per_group_writer_lowest_utilization_single_family_unchanged` — regression bar at the per-group writer.
- Two existing tests updated to pin the family-partition answer:
  - `test_perishable_writer_keeps_both_plans_after_two_polls` (the `[claude-max/fable, openai/astra]` mixed-family scenario).
  - `test_per_group_writer_writes_group_order_hash` (the per-group writer output for a mixed-family body).
  - `test_nested_groups_outer_rotates_inner_picks_lowest_util` — also updated; the inner lowest-utilisation now leads by declared family.

## Verification

```
$ python3 tests/test_routing.py tests/test_classify.py 2>/dev/null | tail -3
  weighted: 50 ultra / 20 max (target ~50/20); affinity pinned minimax-max/m3 -> minimax-max/m3
all routing tests passed

$ for t in tests/test_*.py; do python3 "$t" >/dev/null 2>&1 || echo "FAIL $t"; done
# zero FAIL lines, 392 tests across 18 files

$ python3 tests/render_preview.py
  rendered index.html + fragments (97217 bytes) -> /tmp/switchyard-preview.html
  flat groups preview (10425 bytes) -> /tmp/switchyard-preview-groups.html
  10 plans, 19 models, 6 lanes, 18 quota windows, 4 probes
  group-bearing preview: 3 group header row(s)
```

`conftest.py` blocks non-loopback sockets, so no test reaches a real provider.

## Files

- `switchyard/usage.py` — new `family_partitioned_order` helper (+53 lines).
- `switchyard/portal/app.py` — lane writer applies partition + tier offset (+131 lines).
- `switchyard/portal/groups.py` — per-group writer applies partition + tier offset (+72 lines).
- `switchyard/picker.py` — picker walk re-partitions before the walk (+35 lines).
- `README.md` — `perishable` / `lowest_utilization` strategy docs call out the partition (+13 lines).
- `tests/test_routing.py` — +10 regression tests, two existing assertions updated.
- `tests/test_portal.py` — one existing assertion updated to pin the partition answer.

## Notes

- No schema change. No changes to `config/plans.example.yaml`. The `provider_family` field on `Plan` was already present (used elsewhere in the code base); the partition helper reads it, nothing more.
- No changes to the credential store, `.env`, or `config/plans.yaml`.
- A single-family lane / group produces a byte-identical output to today's behaviour — the regression bar pinned by the two single-family tests and exercised by every existing perishable / lowest_utilization test that doesn't mix families.

Fixes #53


switchyard/portal/app.py
```diff
@@ -244,6 +245,25 @@ async def _recompute_perishable_for_plan(reg, ledger, plan) -> None:
...
+            reg.plan_of(reg.model(ref)).provider_family
+            if reg.model(ref) is not None else None))
+        if previous:
+            prev_refs = [m["ref"] for m in previous["members"]]
+            _, prev_ref_tier = _build_ref_tier(
+                prev_refs, config_refs, family_of)
+            for ref, tier in prev_ref_tier.items():
+                prev = previous_by_ref.get(ref)
+                if prev:
+                    prev["score"] = float(prev["score"]) - tier * _TIER_OFFSET
```
Fledgewing: **[should-fix] Tier offset can get stuck across polls on a config change**

The previous-score strip uses `_build_ref_tier(prev_refs, config_refs, family_of)`, which computes the tier against the **current** config. For the polled plan, the fresh score is recomputed anyway, so this is fine. For other plans in the same lane whose refs are reused from `previous_by_ref`, the strip subtracts the new tier and the writer then re-adds the new tier:

    new_stored = new_tier * 1e9 + (old_stored − new_tier * 1e9)
              = old_stored

So a ref whose tier changed (new family added before its family, its own family reassigned, etc.) keeps its **old** inflated score in the hash until that plan probes again. The picker walk is unaffected — `_visit_order_ranked` re-partitions by current family, never trusting the stored score for ordering — but the capacity board reads `get_lane_order`/`get_group_order` and sorts members by `m["score"]` desc, so the board will show that ref in the wrong bucket until the affected plan's next probe (up to ~2x `probe.interval_seconds`, ~20 min for the shipped 10 min interval). This contradicts the PR's stated rationale ("the reader sorts by score desc, so a per-family tier above the actual score range preserves the partition when the picker / board reads the hash back") — the invariant holds for the picker, but the board view drifts after a config edit.

The fix is small: after the tier-strip loop, store the post-strip value as a *fresh* raw score alongside a per-ref `recompute` flag and let the next plan's probe settle it, or stamp the tier into `score_<ref>` with a per-poll `epoch` so a stale tier is detectable on read. Either way, please add a test that drives the writer with a config that changes the family of a ref between two consecutive polls (e.g., via `replace(...)` on the plan's `provider_family`) and asserts the stored score after the second poll is `1e9 * new_tier + raw`, not the pre-edit stored value.

tests/test_routing.py
```diff
@@ -3558,6 +3577,399 @@ def test_cli_tool_block_parses_strictly_and_normalises_keys():
...
+            "grok", 50.0, reset, window="weekly")
+        await recompute_group_orders(new_reg, ledger)
+        return await ledger.get_group_order(gid, "per-mixed")
+
+    order = run(go())
+    assert order is not None, order
+    refs = [m["ref"] for m in order["members"]]
+    # minimax bucket leads (declared first), xai (grok) trails.
+    # Within the minimax bucket, ultra (more room) outranks max.
+    assert refs == ["minimax-ultra/m3", "minimax-max/m3",
```
Fledgewing, line 3788: **[should-fix] Mixed-family writer tests never assert the 1e9 tier offset is actually stored**

`test_per_group_writer_perishable_mixed_family_ranks_leading_first` (and its `lowest_utilization` sibling) check that `[m["ref"] for m in order["members"]]` returns the family-partition order, but neither inspects `entry["score"]` beyond its positivity. Same for `test_per_group_writer_writes_group_order_hash` in `tests/test_portal.py:803` — only `entry["score"] > 0` is asserted.

The picker is **robust to a missing tier offset**: `_visit_order_ranked` re-partitions by current family (`switchyard/picker.py:484`), so the picker tests would still pass even if the writer regressed and forgot to bake `ref_tier[ref] * _TIER_OFFSET` into the score. The PR's own design rationale explicitly leans on the offset for the board view:

    The reader sorts by score desc, so a per-family tier above the actual
    score range preserves the partition when the picker / board reads the
    hash back. (switchyard/portal/app.py:321–324)

…so a regression that drops the offset would only break the capacity board, which no test exercises. Add a magnitude assertion to at least one mixed-family test in each writer (`app.py` and `groups.py`), e.g.:

    scores = {m["ref"]: m["score"] for m in order["members"]}
    assert scores["minimax-ultra/m3"] >= 1e9   # leading bucket tier
    assert scores["grok/grok-4.6"] < 1e9        # trailing bucket tier = 0
    # and the within-bucket raw score is recoverable
    assert scores["minimax-ultra/m3"] - 1e9 < scores["minimax-max/m3"] - 1e9

That pins both the per-family tier and the within-bucket score-desc ordering, so a refactor that drops the offset (or flips it to trailing-bucket-first) is caught. The same hardening is worth applying to `test_per_group_writer_writes_group_order_hash` and the `single-family_is_byte_identical` lane-writer test — the byte-identity case in particular currently passes for any scrambled order with any positive score, see the sibling comment.

tests/test_routing.py
```diff
@@ -3558,6 +3577,399 @@ def test_cli_tool_block_parses_strictly_and_normalises_keys():
...
+
+    after = run(go())
+    refs = [m["ref"] for m in after["members"]]
+    # Both refs share the None family, so they form one bucket. The
+    # within-bucket input order is the scored-desc / unknown-config
+    # order the writer built -- both are scored, so desc by score. The
+    # higher-room (opus has lower pct_used... wait, both are on the same
+    # plan with the same window facts). They get the same score; the
+    # within-bucket order falls back to the input order. Stable.
+    assert set(refs) == {"claude-max/fable", "claude-max/opus"}, refs
```
Fledgewing, line 3988: **[should-fix] Byte-identity test passes for any permutation and any positive score**

The docstring states the regression bar is "byte-identical to what today's logic would have produced", but the assertions don't enforce it:

```python
assert set(refs) == {"claude-max/fable", "claude-max/opus"}, refs
for entry in after["members"]:
    assert entry["score"] > 0, entry
```

Two consequences:

1. `set(refs)` accepts any order. A regression that scrambled the within-bucket order — say, swapping the entries around `_one_adjacent_swap_partitioned` or returning them in score-desc-without-stable-sort — would still pass. The sibling `test_per_group_writer_lowest_utilization_single_family_unchanged` (line 3936) does this correctly with `assert refs == [...]`; the lane-level test should match.

2. `entry["score"] > 0` accepts any positive value, not the pre-partition raw score. For single-family the spec is `0 * 1e9 + s == s`, so any `s > 0` works — a regression that accidentally applied `1e9 + s` to every ref (e.g., always picking tier=1 in `_build_ref_tier`) would still pass. The test should compute the expected raw score and pin it.

Suggested replacement:

```python
expected_raw = 50.0 / max(1.0, (reset - time.time()) / 3600.0)
assert refs == ["claude-max/fable", "claude-max/opus"], refs
for entry in after["members"]:
    assert entry["score"] == pytest.approx(expected_raw), entry
```

(Or compute it inline without `pytest` if the project is no-pytest, as the rest of the suite.) This is the actual regression bar the spec claims.

switchyard/usage.py
```diff
@@ -653,6 +653,59 @@ def utilization_score(room_pct: float | None, reset_at: float | None = None,
...
+    member order is the input order verbatim; the leading family is the
+    first family to appear in `config_order` (the lane body, or the
+    group's leaf refs in declaration order). Each subsequent family
+    becomes its own bucket in the same first-appearance order. The output
+    concatenates the buckets back-to-back. A single-family input returns
+    an order identical to the input -- the regression bar.
+
+    `family_of` is a `str | None` callable; a None family joins ONE shared
+    "unspecified" bucket keyed (for bucket ordering) as the empty string.
+    The schema makes family either a non-empty string or None, so the
```
Fledgewing: **[nit] Docstring overstates what the schema enforces**

The helper claims:

    The schema makes family either a non-empty string or None, so the
    empty-string key never collides with a real family -- if it ever
    does, the bug is at the schema layer and not here.

But `provider_family` is parsed as `body.get("provider_family")` (`switchyard/models.py:864`) and the dataclass field is `provider_family: str | None = None` (line 273). Nothing rejects an empty string, and nothing rejects whitespace-only strings. A plan configured with `provider_family: ""` (easy to leave by accident in YAML) would compute `key = ""` (since `family is not None` is True for `""`) and silently join the unspecified bucket — exactly the collision the docstring says can't happen.

This is a latent edge case (no shipped plan uses it), so flagging it as a nit rather than a blocker, but the docstring should be tightened to either match reality ("the schema does not enforce a non-empty family; if you want a real family, set it; if you want unspecified, leave it None") or the parser should validate that `provider_family` is either `None` or a non-empty string at load time. The same caveat applies to the new `_partition_with_tier` in `switchyard/portal/app.py` and `switchyard/portal/groups.py`, which inherit the assumption.

Fledgewing:
Reviewed PR #79 at c33d4de. Offline suite green (392 tests / 18 files) per the implementer's report. The partition logic itself is correct at all three sites (lane writer, per-group writer, picker walk), and the picker re-partitions explicitly so a stale hash from a previous deploy is read safely — that handles the cross-version scenario cleanly.

Three findings worth fixing before merge; one nit (not counted in the verdict):

- switchyard/portal/app.py:266 — tier offset can get stuck on a config change. The strip uses the *current* tier, but for OTHER plans' refs (reused from previous) the strip+restore yields `old_stored`, not `1e9 * new_tier + raw`. Picker walk is unaffected (re-partitions); the capacity board view shows the ref in the wrong bucket until its plan polls next. See inline comment for the math and a suggested fix.
- tests/test_routing.py:3779 — mixed-family writer tests check refs order but never inspect `entry["score"]` magnitude. Picker is robust to a missing tier offset (re-partitions), so a writer regression that drops the offset would not be caught by any test. Pin a magnitude assertion on at least one mixed-family writer test in each file.
- tests/test_routing.py:3929 — `test_perishable_lane_writer_single_family_is_byte_identical_to_pre_partition` uses `set(refs)` (no order check) and `entry["score"] > 0` (no value check). The docstring claims byte-identity but the assertions don't enforce it. The sibling `test_per_group_writer_lowest_utilization_single_family_unchanged` does this correctly with `assert refs == [...]`; the lane test should match and pin the exact raw score.

The single nit (not blocking):

- switchyard/usage.py:683 — the helper's docstring claims the schema enforces `provider_family` as non-empty str or None, but `body.get("provider_family")` accepts any string including `""`. Worth either tightening the docstring or rejecting empty strings at parse time.

No blockers. No security concerns. CLAUDE.md repo rules respected (no .env / config/plans.yaml writes, tests before the `__main__` runner and socket-free, no docker from this worktree).

Issue #78, opened 2026-09-22
TEXT_LOST 502 surfaces to the caller on every (mcp_bridge) plan that hits it — should be retried internally, never returned as BadGatewayError


Fledgewing:
## Symptom

After #64 / PR #73, the gateway classifies `cli emitted tokens with no text (prompt_tokens=N, completion_tokens=M)` as `Outcome.TEXT_LOST` with a 10s, non-escalating cooldown and books the charged tokens. **The 502 still reaches the caller as `litellm.BadGatewayError`**, on every `mcp_bridge`-backed plan that hits the shape (currently `opencode-go/glm-5.3-flash` and `opencode-go2/glm-5.3-flash`, which together are the bulk of the `judge` lane's traffic in `config/plans.yaml`).

Live today, `docker logs switchyard-gateway-1`:

```
21:39:33 - LiteLLM Proxy:ERROR: common_request_processing.py:1474 - ... - Error code: 502 -
  {'detail': 'opencode cli emitted tokens with no text (prompt_tokens=150142, completion_tokens=1703)'}.
  Received Model Group=sy.opencode-go2.glm-5.3-flash
21:40:12 - ... - Error code: 502 -
  {'detail': 'opencode cli emitted tokens with no text (prompt_tokens=142717, completion_tokens=36)'}.
  Received Model Group=sy.opencode-go2.glm-5.3-flash
21:40:46 - ... - Error code: 502 -
  {'detail': 'opencode cli emitted tokens with no text (prompt_tokens=150140, completion_tokens=1276)'}.
  Received Model Group=sy.opencode-go2.glm-5.3-flash
```

Counters over the recent tail:
- `grep -c "emitted tokens with no text" → 10` (all on `sy.opencode-go{1,2}.glm-5.3-flash`)
- `grep -c "yielded no parsed answer" → 0` (the WS1 contract string is the only one firing)

The picker *is* working — it skips the freshly-cooled sibling:

```
switchyard: lane=judge -> opencode-go2/glm-5.3-flash [configured + gate headroom 1] tools
  skipped=opencode-go/glm-5.3-flash(pinned, no free slot after 10s),
          opencode-go/glm-5.3-flash(cooled: text_lost)
```

…but both `opencode-go/glm-5.3-flash` and `opencode-go2/glm-5.3-flash` lose text at the same rate, so once `opencode-go` is cooled for text_lost the picker lands on `opencode-go2`, which then hits the same bug, and the caller still fails.

[...]

Issue #13, opened 2026-09-21
Issue: mcp_bridge rebuild leaves the superseded session live — one tool loop can span two live CLI sessions


Fledgewing:
Summary

When a follow-up's tool results match nothing parked in a live session, *continue*followup rebuilds from the request (sidecars/mcp\_bridge/server.py:1032-1044). Two defects turn that one rebuild into churn:

1. Duplicate delivery is misread as a lost session. resolved == 0 fires both for "session is gone" and for "these results were already consumed" (client retry after a dropped socket re-delivers results the first delivery already popped). The comment at server.py:1033-1037 even lists "the GUI replaying an already-delivered result" as a known cause — but the code rebuilds anyway.

2. The rebuild never retires the session it supersedes. resume\_gone\_session → start\_session inserts a fresh session into SESSIONS (server.py:823) while the old one stays parked and live. Nothing calls end\_session on it.

Defect 2 breaks the invariant "one caller tool loop ↔ at most one live mcp\_bridge session". Once two sessions of the same loop are live, every later follow-up is ambiguous.

Evidence (opencode-go2, today 21:47–21:56 UTC, session ses\_f3a0e9df4ffeybWNKuk9tn1lzV)

follow-up tool results span 2 live mcp_bridge sessions (['e648f2b8…', '986e8882…']); resolving against e648… and dropping the others' tool results

live mcp\_bridge session e648… had no parked tool call matching the delivered ids; rebuilding from the request so the caller is not trapped

POST /v1/chat/completions 502 Bad Gateway        ← "opencode cli yielded no parsed answer" (cli\_bridge server.py:634 → 791)

… span-2 warning repeats, then:

POST /v1/chat/completions 503 Service Unavailable ← "no slot freed … within 300s to resume"

The client retried throughout (242 "socket connection was closed" in the client log today is the retry source), producing the 4-min 429 pinned to opencode-go2 … no free slot loop on the gateway.

Suggested fix

- On the rebuild path, end\_session() (server.py:578) the superseded session before/when starting the replacement — both in *continue*followup's resolved == 0 branch and for the dropped sessions in handle\_followup's span&gt;1 branch (server.py:966-984), since a session whose results were dropped can never be continued coherently.

- Track recently-resolved tool\_call\_ids per session (small ring with TTL; Session.batch\_ids is close already). If every delivered id was already resolved, return the stored turn result / a no-op continuation instead of rebuilding — "client never sees an error" without paying a rebuild.

[...]

PR #82, opened 2026-09-22
mcp_bridge: release the gate slot when a turn ends in an error result (#80)

### LINKED ISSUE #80: mcp_bridge leaks a gate slot on every error-result turn — gateway cap has to drop to 1 to recover
## Symptom

A new request to `sy.opencode-go2.glm-5.3-flash` (or `sy.opencode-go.glm-5.3-flash`) hits `RateLimitError: 429 - {'detail': 'sidecar at capacity (4)'}` even though the gateway's picker saw a free slot. Repeated on every following request until the gateway's concurrency learner drives `learned[global]` cap all the way down to 1. Live in production right now:

```
gateway-1  | switchyard: lane=judge -> opencode-go2/glm-5.3-flash [configured + gate headroom 1] tools skipped=opencode-go/glm-5.3-flash(lease unusable),opencode-go/glm-5.3-flash(cooled: concurrency)
gateway-1  | switchyard: plan=opencode-go2 refused on connection limit at 3 — learned cap now 1
gateway-1  | 21:49:37 - LiteLLM Proxy:ERROR: ... - Error code: 429 - {'detail': 'sidecar at capacity (4)'}. Received Model Group=sy.opencode-go2.glm-5.3-flash
```

The gateway cap dropped from 3 to 1 on every sidecar rejection (the `note_rejection(at_cap=3)` halving pattern), so the plan is now permanently limited to one concurrent request until the gate is fully cleared. Opencode-go is in the same shape one step earlier.

## Root cause

`sidecarcars/mcp_bridge/server.py:1311` (`handle_fresh`) acquires a slot on `cli_bridge._gate` before calling `start_session`. `start_session` (line 1388-1399) does the normal work:

```python
result = await await_turn(session, request)
response = render_turn(session, result, body.get("model"))     # raises on error result
session.last_response = response
if result["type"] != "tool_calls":
    await end_session(session)                                  # only here on success/final
else:
    await park_session(session)                                # only here on tool_calls
```

`render_turn` raises `HTTPException` for `result["type"] == "error"` (line 1241-1243). Every CLI subprocess outcome that produced an error (TEXT_LOST, auth failure, non-zero exit, parse failure, timeout, ...) reaches this branch. The exception propagates up; `handle_fresh` catches it (line 1320) only to clean up `img_dir` and re-raise. Neither `end_session` nor `park_session` runs, so the gate slot stays held by a Session that the sidecar still has in `SESSIONS` (`holds_slot=True`, `dead=False`).

Same shape on `_continue_followup` (unpark→await_turn→render_turn, line 1710-1723) and on `resume_gone_session` (acquire_resume_slot→start_session, line 1543-1554): both can leak the slot the same way on error results.

The companion `cli_bridge` path is correct: `_handle_chat` wraps the work in `try/finally` that always calls `_gate.release()` (cli_bridge/server.py:1729-1732). The mcp_bridge path was never given the same try/finally.

## Why the gateway learns the wrong cap

The gateway picker only ever claimed at most 3 slots (`max_parallel=4 - gate_headroom=1`). On a normal completion, `async_post_call_success_hook`/`async_post_call_failure_hook` release the plan-slot entry in Redis (`sy:inflight:opencode-go2`). So the gateway thinks "0 in flight, 3 free" and picks again. The sidecar, however, still has the leaked Session holding its gate slot. After 4 leaks the gate is at its physical cap and the next acquire returns False — the 429 the caller saw.

`at_cap = int(ctx.get("cap") or plan.max_parallel)` (hooks.py:1159). `ctx["cap"]` is the gateway's picker cap (3 after headroom), not the sidecar's physical cap (4). `ConcurrencyLearner.note_rejection` does `new = max(floor, int(at_concurrency * decrease_factor))`, so `int(3 * 0.5) = 1`. The plan's learned cap drops from 3 to 1, the picker admits at most 1, but the gate still has 4 zombies. The next release is the idle reaper (1800s TTL) — far too long to wait through.

## Reproduction (live state right now)

```
docker compose exec opencode-go2-sidecar curl -s http://localhost:8085/health | jq '{in_flight, sessions, awaiting_followup}'
{
  "in_flight": 4,
  "sessions": 4,
  "awaiting_followup": 0
}

docker compose exec redis redis-cli -n 1 ZCARD sy:inflight:opencode-go2
0

docker compose exec redis redis-cli -n 1 HGETALL sy:learn:opencode-go2:global
cap 1
rejections 2
```

- Sidecar: gate count 4, 4 sessions in `SESSIONS`, none parked, **no CLI subprocesses** (`ps` shows only uvicorn).
- Gateway: 0 entries in `sy:inflight:opencode-go2`. The picker will admit any request up to its learned cap (1 now).
- The 4 zombie sessions are from 21:39:12 / 21:40:05 / 21:40:29 / 21:41:23 UTC. Each was a fresh `start_session` whose CLI subprocess returned a TEXT_LOST 502 (`opencode cli emitted tokens with no text (prompt_tokens=…, completion_tokens=…)` — the issue #78 contract shape). `render_turn` raised, `end_session` never ran, gate slot stayed held.

The four workdirs (`/tmp/mcpb-*`) are still on disk; the Session objects are still in memory. Nothing in the existing lifecycle (idle reaper, `end_session`, `park_session`, `reap_session`) will free them — they need the `error` branch to release, and the `error` branch doesn't reach `end_session`.

## Fix

In `mcp_bridge/server.py:start_session`, wrap the post-`SESSIONS[…]=session` block in `try/finally` and always run `end_session(session)` (or equivalent: `session.holds_slot = False; await cli_bridge._gate.release(); SESSIONS.pop(...)`) on the error path. The early failure path at line 1372-1383 already does this manually for `build_argv` failures; the same release needs to happen for `render_turn` raising. `cli_bridge._handle_chat`'s `try/finally` (cli_bridge/server.py:1729-1732) is the model.

For `_continue_followup` (line 1710-1723), the same `try/finally` wraps the `unpark_session → await await_turn → render_turn → end_session/park_session` block, and releases the slot on the error path. For `resume_gone_session` (line 1543-1554), the gate slot from `acquire_resume_slot` is released by the same `end_session` call inside the wrapped `start_session`, so wrapping `start_session` covers it too — no separate fix needed at the rebuild-path site.

Cleanup also needs to remove the zombie Session from `SESSIONS` (the in-memory dict that `/health` reports as `sessions`), or `/health` will keep showing the leak long after the gate is fixed. `end_session` already does that (`SESSIONS.pop(session.id, None)`), so reusing it is the right primitive.

## Test plan

In `tests/test_mcp_bridge.py`, alongside the existing gate accounting tests at line 455-484:

- `test_error_result_releases_gate_slot_and_removes_session` — drive `handle_fresh` with a stubbed `start_session` (or stub `run_session` to resolve_final `{"type": "error", "status": 502, "detail": "x"}`) and assert `cli_bridge._gate.in_flight == 0` and `session_id not in server.SESSIONS` after the HTTPException propagates.
- `test_text_lost_does_not_leak_a_slot` — same shape, with the WS1 contract detail `f"{PROVIDER} cli emitted tokens with no text (prompt_tokens=N, completion_tokens=M)"` so the regression test guards the exact shape that's leaking today.
- `test_followup_error_result_does_not_leak` — drive `_continue_followup` with an `unpark_session` against a fresh gate, then `resolve_final({"type": "error", ...})`; assert the gate is back to its baseline.
- `test_rebuild_error_does_not_leak` — drive `resume_gone_session` with the same error outcome; assert `cli_bridge._gate.in_flight == 0`.

The existing `test_unpark_session_returns_503_quickly_when_gate_is_full` (line 1380+) and `test_resume_gone_returns_503_quickly_when_gate_is_full` (line 1423+) only exercise the acquire-failed path. Adding the four above closes the leak in the same test file that already documents the slot accounting contract.

## Evidence appendix

```
# Sidecar state (the live symptom)
docker compose exec opencode-go2-sidecar curl -s http://localhost:8085/health | jq '{in_flight, sessions, awaiting_followup}'
# { "in_flight": 4, "sessions": 4, "awaiting_followup": 0 }

# Gateway state (it thinks the plan is empty)
docker compose exec redis redis-cli -n 1 ZCARD sy:inflight:opencode-go2
# 0

# No CLI subprocesses — the zombies are real
docker compose exec opencode-go2-sidecar ls /proc | grep -E '^[0-9]+$' | wc -l
# 3   (shell, my exec, uvicorn)

# The four leaked sessions, with their creation times
docker compose exec opencode-go2-sidecar stat -c '%y %n' /tmp/mcpb-* 2>&1 | sort
# 2026-09-22 21:39:12 /tmp/mcpb-0dd41fdd-oez3pg83
# 2026-09-22 21:40:05 /tmp/mcpb-b24e0f47-imqav13k
# 2026-09-22 21:40:29 /tmp/mcpb-ffb23771-uaysw8kp
# 2026-09-22 21:41:23 /tmp/mcpb-1935b079-lvm0kmb6

# The 502s that triggered each leak (gateway log)
docker compose logs gateway --tail 4000 | grep -E "21:3[89]|21:4[0-1].*opencode-go2.*emitted tokens"
# 21:39:33 ... opencode-go2 ... emitted tokens with no text (prompt_tokens=150142, completion_tokens=1703)
# 21:40:12 ... opencode-go2 ... emitted tokens with no text (prompt_tokens=142717, completion_tokens=36)
# 21:40:46 ... opencode-go2 ... emitted tokens with no text (prompt_tokens=150140, completion_tokens=1276)
# 21:41:31 ... opencode-go2 ... emitted tokens with no text (prompt_tokens=150143, completion_tokens=36)

# The cascading learned-cap collapse
docker compose logs gateway --tail 4000 | grep "opencode-go2 refused on connection limit"
# plan=opencode-go2 refused on connection limit at 3 — learned cap now 1
# plan=opencode-go2 refused on connection limit at 1 — learned cap now 1
```

Relates to #78 (same TEXT_LOST shape, but the caller-visible symptom). This issue is the underlying state-machine bug that makes #78's plan eventually refuse to admit any request at all — fixing the leak is also what stops #78's retry from being blocked by its own plan's ghost capacity.

Fledgewing:
## Problem

`render_turn` raises `HTTPException` for every CLI-failure turn
(TEXT_LOST, auth failure, non-zero exit, parse failure, timeout, 413,
429). That exception escaped `start_session` and `_continue_followup`
**without** running `end_session`, so the `Session` stayed in
`SESSIONS` holding its gate slot until the 1800 s idle reaper came
round. Live evidence from the issue:

- gate `in_flight: 4` (full), 4 zombie sessions in `SESSIONS`
  from a string of TEXT_LOST 502s
- every subsequent request 429'd `sidecar at capacity (4)`
- the gateway's `ConcurrencyLearner.note_rejection(at_cap=3)` halved
  the plan's learned cap (3 -> 1), so even after the leaks aged out
  the sidecar stayed throttled

`resume_gone_session` (the rebuild path) inherited the same bug
through `start_session`.

## Fix

Wrap the post-acquisition block in `start_session` (after
`SESSIONS[session_id] = session` / `new_turn()`) and in
`_continue_followup` (after `unpark_session`):

```python
try:
    ...the existing block...
except BaseException:
    await end_session(session)
    raise
```

Deliberately **not** a bare `finally: end_session` -- parked
`tool_calls` sessions hold no slot by design and must survive the
success path; that invariant is already pinned by
`test_a_parked_session_holds_no_concurrency_slot` and now also by the
`tool_calls`-success branches of the new tests, which would fail with
the naive `finally`.

`BaseException` (not `Exception`) catches the cancel case too; the
`unpark_session` 503-on-saturated-gate path is intentionally left
**outside** the new try, because that failure leaves the session
parked for the gateway to retry (issue #13 follow-up semantics).

`resume_gone_session` is covered transitively via `start_session`,
plus a comment at the rebuild handoff point naming the new teardown
contract. The unpark's pinned 503 path and the 1800 s reaper are
untouched.

## Tests

Added four regression tests in `tests/test_mcp_bridge.py`, inserted
before the `__main__` runner (51 -> 55 passed):

- `test_error_result_releases_gate_slot_and_removes_session` --
  fresh `handle_fresh` path with a 502 error dict; asserts the
  HTTPException 502 propagates, gate `in_flight == 0`, session gone
  from `SESSIONS`, dead, `not holds_slot`.
- `test_text_lost_does_not_leak_a_slot` -- same harness, but the
  detail string comes from `cli_bridge.format_no_text_detail(...)`
  and is checked **byte-for-byte** against the gateway's `_NO_TEXT`
  regex classification (issue #64 shape).
- `test_followup_error_result_does_not_leak` -- hand-built parked
  session in `awaiting_followup` / `holds_slot=False`, stubbed
  `await_turn` returns a 502 error mid-tool-loop. Covers the
  `_continue_followup` path with the same invariants.
- `test_rebuild_error_does_not_leak` -- `resume_gone_session` with a
  fresh gate (so `acquire_resume_slot` returns True), rebuilt CLI
  raises 502; the rebuild path inherits the teardown contract.

Full test-suite run (excluding the four pre-existing
`ModuleNotFoundError: litellm` failures in
`test_messages_hooks.py`, `test_routing.py`,
`test_session_injection.py`, `test_verdict.py`, which fail identically
on baseline without the fix) is otherwise clean.

## Out of scope

- gateway `hooks.py` / `policy.py` cap-learning: downstream symptom,
  self-corrects once the sidecar stops 429ing the picker.
- issue #78's retry behaviour: related caller-visible TEXT_LOST
  shape, separate fix.

## Deploy note

`switchyard-sidecar` code is baked into the sidecar image, not
mounted. After merge, rebuild + redeploy happens from the **main
checkout** (`scripts/apply.sh --build`), never from a worktree --
this worktree's `switchyard/*.py` (none touched) does not need a
rebuild, only the sidecar image does.

Fixes #80


sidecars/mcp_bridge/server.py
```diff
@@ -1384,19 +1384,34 @@ async def start_session(body: dict, mcp_tools: list[dict], prompt: str,
...
+        # Both must tear the session down: otherwise the session stays in
+        # SESSIONS holding its gate slot until the 1800 s idle reaper comes
+        # round, and four such leaks physically fill the sidecar gate (issue
+        # #80). BaseException (not Exception) catches the cancel too;
+        # end_session's `if session.dead: return` makes the call idempotent
+        # against the 499-already-reaped path that await_turn can raise when
+        # the caller has hung up. NOT a bare `finally: end_session` -- parked
+        # sessions hold no slot by design and must survive the tool_calls
+        # success path.
         await end_session(session)
```
Fledgewing: **[should-fix] Cleanup itself can be re-cancelled, leaving the slot leaked**

The new `except BaseException` block awaits `end_session(session)`, which itself awaits `cli_bridge._gate.release()`. If the request is cancelled a *second* time during that cleanup -- a server-shutdown re-cancel hitting the same task, a Starlette disconnect followed by FastAPI's lifespan shutdown, anything that calls `task.cancel()` twice -- `await gate.release()` raises `CancelledError` mid-cleanup: the slot is decremented (or not, depending on where the cancel hit) and the partial teardown is lost. That is the same leak shape as before, just shifted one frame out.

The PR description claims "BaseException (not Exception) catches the cancel too", but that only catches the *first* cancellation. The cleanup needs to be defensive against a second one arriving while it's running:

```python
except BaseException:
    with contextlib.suppress(BaseException):
        await end_session(session)
    raise
```

(or wrap in `try / except BaseException: log...; pass` for visibility). Same shape applies to the mirror block at line 1755 in `_continue_followup`.

tests/test_mcp_bridge.py
```diff
@@ -1502,6 +1502,351 @@ async def fake_start(body, mcp_tools, prompt, system, model, request,
...
+        captured_session["session"] = session
+
+        # _continue_followup awaits `session.new_turn()` to install a new
+        # turn_future. In production the existing run_session task resolves
+        # that future when the CLI emits its next answer; here we stub
+        # await_turn to return the error dict directly, but the await on
+        # new_turn itself still has to be unblocked or the test hangs.
+        # Override new_turn on this session only so it returns an
+        # already-resolved future -- await_turn (stubbed) reads
+        # session.turn_future but ignores it.
```
Fledgewing: **[nit] Test comment misstates the rationale for the `new_turn` patch**

The comment says: `await_turn (stubbed) reads session.turn_future but ignores it`. The stub `error_turn(sess, request)` does not read `sess.turn_future` at all -- it just `return ERROR`s, sidestepping the whole `await turn` dance in the real `await_turn`.

The actual reason the patch is required is correctly given in the paragraph above: `await session.new_turn()` would otherwise block on an unresolved future because `run_session` is never started in this scenario. The misleading sentence confuses that -- a future reader sees "we patch new_turn because the await_turn stub ignores turn_future" and goes looking for a stub that ignores turn_future.

Suggest trimming the last clause and rephrasing, e.g.:

```python
# Override new_turn on this session only so it returns an
# already-resolved future -- otherwise `await session.new_turn()`
# would hang (run_session is never started in this scenario).
```

Fledgewing:
Reviewed 2 findings: 1 should-fix, 1 nit — see inline comments at cc53379.

The fix correctly closes the slot leak issue #80 identified. The new `except BaseException` guards in `start_session` (line 1401) and `_continue_followup` (line 1745) catch every CLI failure path that `render_turn` raises for, and `end_session`'s `if session.dead: return` guard keeps the cleanup idempotent against the 499-already-reaped path. The unpark 503 path is correctly left outside the try (line 1731), and the parked-session survival is preserved by the `if result["type"] != "tool_calls"` branch inside the try (not in `finally`); the `tool_calls` success path still falls through to `park_session` as before. The `resume_gone_session` rebuild path is covered transitively via `start_session`, with a comment marking the handoff. All four new regression tests run (55/55 pass under `/tmp/opencode/venv/bin/python3 tests/test_mcp_bridge.py`), assert the right invariants (`gate.in_flight == 0`, `session.id not in SESSIONS`, `session.dead`, `not session.holds_slot`), and run before the `__main__` runner. Diff scope is contained to `sidecars/mcp_bridge/server.py` + `tests/test_mcp_bridge.py`; no touches to `hooks.py`, `policy.py`, `cli_bridge`, or config.

The should-fix concerns the cleanup itself being cancellation-safe under a *second* cancel arriving during `end_session`; the nit is a misleading comment in the follow-up test.

sidecars/mcp_bridge/server.py
Fledgewing: Fixed in 3deecac: wrapped the `await end_session(session)` in `with contextlib.suppress(BaseException):` in both `start_session` and `_continue_followup`, so a second `CancelledError` interrupting the cleanup no longer strands the slot. The `raise` outside the `with` still re-raises the original exception, and the 1800 s reaper remains the last-resort backstop if even the suppressed path did not run to completion. Thanks for catching the second-cancel gap.

tests/test_mcp_bridge.py
Fledgewing: Fixed in 3deecac: rewrote the comment in `test_followup_error_result_does_not_leak` to read "...otherwise `await session.new_turn()` would hang (run_session is never started in this scenario)", dropping the misleading clause about the stub reading `turn_future`. The `run_session`-never-started point was already made in the paragraph above; the misleading sentence just confused the rationale by referencing a stub that does not actually inspect `turn_future`.

PR #84, opened 2026-09-22
Board: show $/session (proxy for $/task) per model alongside $/MTok

### LINKED ISSUE #76: Board: show $/session (proxy for $/task) per model alongside $/MTok — building block for cost-based routing
### Problem

The portal's economics column shows `$/MTok` per model (`switchyard/portal/templates/_plans.html:133`), computed by `model_effective_cost_per_mtok` (`switchyard/usage.py:763`). That's the right number for "is this subscription worth renewing" — a $132 plan that delivered 40M tokens cost $3.30/Mtok, the same plan at 400M cost $0.33. But it is the wrong number for "what does one task cost me on this model," which is the number cost-based routing needs.

Two models can both report `$/MTok = $0.50` and yet be wildly different in absolute spend per user interaction:

- A $200 plan that handled 5 long multi-turn sessions: 40M tokens @ $0.50/Mtok = $20 spent on the work, but $40 per session.
- A $20 plan that handled 50 short sessions: 40M tokens @ $0.50/Mtok = $20 spent on the work, but $0.40 per session.

Same MTok rate, **two orders of magnitude** apart in unit economics. `/MTok` flattens that into one number, and the operator cannot see which model is the better deal for the kind of work their agents actually do.

The eventual goal is **cost-based routing**: classify a request by task shape and route it to the model with the lowest expected cost for that shape. Before any of that, we need the building block — a per-model cost-per-task proxy we can show, read, and eventually feed the picker. This issue ships just that building block.

### Proposal

Add a second economics cell to each model row on the portal: **`$/session`**, alongside the existing `$/MTok`. Session is the proxy for task because we don't have a clean task boundary on the wire — the closest signal is "one client conversation thread," which is exactly what `session.derive()` (`switchyard/session.py:49`) already produces for affinity. Same source of truth, same edge cases, one less invented concept.

Surface shape:

```
month_tokens         eff cost
42M   $21            $0.50/MTok    $3.85/session
```

Same threshold gating as `$/MTok` today: a model with fewer than N unique sessions in the window shows `—`, not a misleading number built on one or two data points.

### Implementation sketch

Mirrors `model_effective_cost_per_mtok` (`switchyard/usage.py:763`) and `model_overview` (`switchyard/usage.py:159`) so the addition is additive — existing call sites and config are unchanged.

1. **Count unique sessions per model per bucket window.** Today the ledger tracks `cost`, `prompt_tokens`, `completion_tokens` in `K_M_HOUR` / `K_M_DAY` hashes (`switchyard/usage.py:177-184`). Add a HyperLogLog per (plan, model, hour|day) keyed on the session id derived in step 2:

   ```
   sy:sessionh:{plan}:{model}:{YYYY-MM-DDTHH}    HLL of session ids seen that hour
   sy:sessiond:{plan}:{model}:{YYYY-MM-DD}       HLL of session ids seen that day
   ```

   `PFADD` at the same point the bucket writes happen today (the success-logging event, alongside the existing `cost` / token accounting — both `async_log_success_event` and `async_post_call_success_hook` paths land in `_finish_success`). `PFCOUNT` reads, ~0.81% error, standard error bound for an economic display. Cheaper than storing the set and bounded by Redis's native HLL encoding (~12 KB per key regardless of cardinality). A session is only counted on the **first** request that lands on a given model inside the window — `PFADD` is naturally idempotent on the same value, so re-leases on the same model are free.

2. **Session id source.** Use `session.derive(data, key_hash)` from `switchyard/session.py:49` — exactly what the picker already uses for affinity. Stamped on the request metadata in `hooks.py:363` as `meta[META_KEY]["session"]`, so the success-logging hook can read it from the same `ctx` it already pulls plan/model/rid from. No new header parsing, no new client contract. **Caveat called out below**: derived-from-prefix fingerprints (`fp:` ids) change when the conversation prefix changes, so a single client that rephrases the prefix between turns can drift to a new "session" by our count. Same drift the affinity table already has, so the number is at least no worse than what we route on today.

3. **New helper in `switchyard/usage.py`**, parallel to `model_effective_cost_per_mtok`:

   ```python
   def model_effective_cost_per_session(
       plan: Plan, n_sessions: int, model_cost: float, plan_sessions: int,
   ) -> float | None:
       """Per-model effective $/session.

       Same gating rules as model_effective_cost_per_mtok, with the threshold
       being a session count instead of a token count -- spend is the same,
       only the denominator changes.
       """
   ```

   Subscription plans: allocate `plan.monthly_cost` across models by session share. Metered plans: spend is the model's own metered cost. Return `None` below the threshold so the template can render `—`. Pick the threshold by experimentation — start with `n_sessions >= 50` for subscription plans (high enough that a 1-session anomaly can't dominate, low enough that a week of real use crosses it) and `n_sessions >= 100` for metered plans where each session is a real unit of billing. Revisit once we have data.

4. **Extend `model_overview`** (`switchyard/usage.py:159`) to read both HLL keys in the same pipelined round trip and return `n_sessions` alongside `month_tokens` / `month_cost`. The same pipeline pattern that already batches 31 day-keys + 3 hour-keys per model — adding 32 more reads per model is a non-event.

5. **Portal wiring.** `portal/app.py:500` already builds the `eff_cost` field; add a sibling `eff_cost_per_session` using the new helper. Render a second line under the existing `$/MTok` cell in `_plans.html:130-137`, or extend the cell to two lines — whichever reads better on a narrow model column. Make it opt-in via a `Settings.show_cost_per_session` flag (default `true`) so an operator who finds the column noisy can hide it without code changes.

6. **Tests** in `tests/test_usage.py` next to `test_effective_cost_per_mtok_unchanged` (`tests/test_usage.py:289`): the threshold gate, the subscription-vs-metered allocation rule, and the `None` returns for cold data. Plus an HLL-level test that confirms `PFADD` is idempotent on the same session id across re-leases on the same model (the "one session = one count, not N" invariant).

### What this is NOT

- **Not cost-based routing.** This is the metric; the picker doesn't read it. The follow-up that uses it (likely a `lowest_session_cost` strategy or a classifier-driven default) is a separate issue — this one only ships the number and the read path.
- **Not a per-request cost prediction.** $/session is a backward-looking economic aggregate. It says "on average, what did a session cost on this model," not "this next session will cost $X." Per-request prediction needs the classifier that cost-based routing is going to require anyway, and that's a different building block.
- **Not a replacement for $/MTok.** They measure different things. `/MTok` answers renewal math; `/session` answers workload-shape math. Both should stay.

### Resolved

- **Session as proxy for task.** A task boundary isn't on the wire; session is what the picker already treats as "one unit of work." Until a client surfaces a task id, we use the same signal and inherit the same edge cases — including the prefix-fingerprint drift caveat above. Documenting the proxy explicitly is more honest than pretending we measure tasks.
- **HLL, not SET.** Cardinality on a Redis SET grows linearly with traffic; over a month on a busy plan that's millions of session ids stored. HLL is bounded (~12 KB), within ~0.81% of the true count, and ships with Redis. The display is rounded, so the error is invisible.
- **Count on success only.** A session that landed on a model but failed every request still occupies the lease table and pays for retries. Counting it would understate `/session` for flaky models. Counting it wouldn't. We pick "would" — the cost is real, the failure is the model's problem, and `/session` should reflect what the operator is paying.
- **Per-month window.** Hour/day buckets already exist for tokens; sessions aggregate the same way. `/session` is a monthly number to match `/MTok`.

### Still open

- Threshold for "enough sessions to mean anything." Proposal starts at 50 (subscription) / 100 (metered). Need a couple of weeks of real traffic to confirm; easy to revisit.
- Whether to expose a per-CLI breakdown (`$/session` split by Claude Code vs Cursor vs Codex) the same way `caller_env` already stamps the request. Useful but separate; can land in a follow-up.
- Whether the portal cell should be sortable. Cheap to add once the column exists.

Fledgewing:
## Why

`$/MTok` answers "is this subscription worth renewing" but is the wrong
number for "what does one task cost me on this model" — the number
cost-based routing will need. Two models at the same $/MTok can differ by
two orders of magnitude in spend per session (long multi-turn vs short).
This ships the missing metric: per-model **$/session**, with session (one
client conversation thread, `session.derive()`) as the declared proxy for
task. Closes #76.

## What

- **Counting** (`switchyard/usage.py`): new model-scoped monthly HLL
  `K_M_SMONTH`; `Ledger.record` PFADDs the session id on success events
  only (idempotent per session — re-leases don't double count) with the
  same TTL pattern as the day buckets; `model_overview` PFCOUNTs it in the
  existing pipeline and returns `n_sessions` per model. One month key
  instead of hour+day keys avoids midnight-boundary double counts; ~12 KB
  per (plan, model, month).
- **Metric** (`switchyard/usage.py`): `model_effective_cost_per_session`,
  parallel to `model_effective_cost_per_mtok`. Metered: model's own
  metered spend / its sessions. Subscription: `monthly_cost /
  plan_sessions` (session-share allocation cancels). Gated: `—` below
  50 sessions (subscription) / 100 (metered), matching the issue's
  proposal; thresholds are module constants like the existing /MTok gate.
- **Gateway** (`switchyard/hooks.py`): success path passes the already-
  stamped `ctx["session"]` to the ledger (one line; no new client
  contract, no re-stamping).
- **Portal** (`switchyard/portal/app.py`, `templates/_plans.html`): model
  rows carry `n_sessions` + `eff_cost_session`; the board renders a second
  `$/session` line under `$/MTok`, `—` when gated. New
  `show_cost_per_session` setting (default true) hides it
  (`config/plans.example.yaml` updated; live `config/plans.yaml` is the
  operator's and picks the key up with its default).
- **Tests**: `tests/fake_redis.py` gains pfadd/pfcount (set-membership
  semantics); six new tests in `tests/test_usage.py` (threshold gates,
  metered vs subscription allocation, cold data, zero spend, and an
  integration test proving one session counts once across re-leases).
  `tests/render_preview.py` fixture rows updated for the new key.

## Not in scope (per the issue)

No cost-based routing (the picker doesn't read `n_sessions`), no
per-request cost prediction, `$/MTok` stays.

## Verification

Offline suite green (`tests/test_usage.py` 27 → 33 tests, `test_portal.py`,
`test_session_injection.py`, `render_preview.py`, full
`for t in tests/test_*.py` loop). No real providers, no live Redis.
Note for deploy: `switchyard/*.py` is baked into the images — the usual
`scripts/apply.sh --build` from the main checkout ships it.


switchyard/portal/app.py
```diff
@@ -473,6 +474,11 @@ async def collect_plans() -> list[dict]:
         model_overview = await ledger.model_overview(
             plan.key, model_refs, month
         ) if model_refs else {}
+        # Total sessions-this-month for the plan, summed across every model
+        # ref. The session-rate helper needs the PLAN's session count (not
+        # the model's) for the subscription share-cancels calculation.
+        plan_sessions = sum(ov.get("n_sessions", 0)
```
Fledgewing: **[should-fix] `plan_sessions` over-counts distinct sessions; the subscription rate understates**

`plan_sessions = sum(ov.get("n_sessions", 0) for ov in model_overview.values())` adds per-model HLL cardinalities rather than taking the union. A session that lands on multiple models within one plan — e.g. mid-loop spillover from `claude-max/opus` to `claude-max/fable` on the same plan, or any lane that lists several models of one plan (`nest-demo` lists `claude-max/opus` and `claude-max/fable` together) — counts once per model it touched, so `plan_sessions` over-counts the true distinct-session count for the plan by however many models that session spanned.

Because `model_effective_cost_per_session`'s subscription branch collapses to `monthly_cost / plan_sessions` (the two `n_sessions` cancel), an over-counted denominator makes the published `$/session` for a subscription understate the truth by the same factor. The metered branch is unaffected. The `plan_sessions < 50` gate is unaffected (the sum is an upper bound on the union, so it gates at least as permissively as it should).

The comment a few lines up calls this "the PLAN's session count (not the model's)"; a reader would reasonably expect the plan's distinct cardinality, not the per-model sum.

Fix without adding a new HLL: Redis's `PFCOUNT` accepts multiple keys and returns the cardinality of the union in one call (`PFCOUNT k1 k2 k3 ...`). Queue one `pipe.pfcount(*smonth_keys_for_plan)` after the hash reads in `model_overview`, surface the result as a single plan-level `n_sessions` (or keep per-model + add a plan-level one), and drop the sum in `collect_plans`.

switchyard/usage.py
```diff
@@ -791,3 +817,44 @@ def model_effective_cost_per_mtok(
...
+    if n_sessions <= 0:
+        return None
+    if plan.metered:
+        if n_sessions < SESSION_MIN_METERED:
+            return None
+        spend = model_cost
+    else:
+        if plan_sessions < SESSION_MIN_SUBSCRIPTION:
+            return None
+        spend = (plan.monthly_cost or 0.0) * n_sessions / plan_sessions
```
Fledgewing: **[nit] Subscription allocation does a round-trip cancel that could just be a single division**

```python
spend = (plan.monthly_cost or 0.0) * n_sessions / plan_sessions
...
return round(spend / n_sessions, 4)
```

`n_sessions` multiplies and then divides out, leaving `monthly_cost / plan_sessions` — which the docstring already states is the intended rate. The form preserves the "fee × share / share" narrative but adds an extra multiply + divide, forces a reader to verify the cancellation, and (marginally) opens the door to precision drift.

Not a correctness bug — just `return round((plan.monthly_cost or 0.0) / plan_sessions, 4)` for the subscription branch, matching the docstring verbatim. The metered branch is fine as-is.

switchyard/usage.py
```diff
@@ -200,6 +222,10 @@ async def model_overview(
             else:
                 row["month_cost"] += b["cost"]
                 row["month_tokens"] += b["prompt_tokens"] + b["completion_tokens"]
+        # The trailing PFCOUNTs come after every hash read in queue order,
+        # so the slice starts at len(reads).
+        for ref, count in zip(model_refs, raw[len(reads):]):
```
Fledgewing: **[nit] Positional slice `raw[len(reads):]` contradicts the existing "never positional arithmetic" rule**

`model_overview`'s body has long warned against this pattern: "the Ordered read plan: results come back in queue order, so aggregation is a zip against this list, never positional arithmetic on slices." The new PFCOUNT decoding uses exactly that positional slice:

```python
# The trailing PFCOUNTs come after every hash read in queue order,
# so the slice starts at len(reads).
for ref, count in zip(model_refs, raw[len(reads):]):
    out[ref]["n_sessions"] = int(count or 0)
```

The inline comment justifies it (PFCOUNTs are always last in queue order), but the style is inconsistent with the hash-read loop a few lines above, which keeps its own parallel `reads` list and zips against it. If `model_overview` ever grows a third operation appended after the PFCOUNTs (e.g. another HLL for a different window), this slice silently mis-attributes results.

Build a parallel list when the PFCOUNTs are queued, and zip against it the same way the hash reads do:

```python
smonth_reads: list[tuple[str, str]] = []
for ref in model_refs:
    key = K_M_SMONTH.format(plan=plan_key, model=ref, period=month)
    pipe.pfcount(key)
    smonth_reads.append((ref, key))
...
for (ref, _), count in zip(smonth_reads, raw[len(reads):]):
    out[ref]["n_sessions"] = int(count or 0)
```

Same slice, but the queue-order invariant is now explicit in the data structure instead of living only in a comment that a refactor could delete without anyone noticing.

tests/test_usage.py
```diff
@@ -608,6 +610,129 @@ def test_lane_order_hash_still_stale_when_no_plan_in_grace():
...
+
+    metered = fake_plan(monthly_cost=0.0, metered=True)
+    assert model_effective_cost_per_session(metered, 0, 0.10, 200) is None
+
+    # Subscription with monthly_cost=0 and live sessions -> 0.0, not None.
+    free = fake_plan(monthly_cost=0.0, metered=False)
+    assert model_effective_cost_per_session(free, 200, 0.0, 200) == 0.0
+
+
+def test_model_overview_counts_unique_sessions():
```
Fledgewing, line 698: **[nit] No test pins "failed=True + session present does NOT PFADD"**

`record()`'s PFADD gate is `if model and session and not failed:` — `failed=True` is supposed to skip the HLL even when both `model` and `session` are set. `test_model_overview_counts_unique_sessions` exercises two of the three input combinations that matter here:

- success path with `session="session-A"` re-recorded 5x → PFCOUNT = 1 ✓
- success path with `session` omitted entirely → no HLL key created ✓

It does NOT cover the third branch: a record that has BOTH `model` and `session` set but `failed=True`. That branch is what protects the rate from being polluted by a session that landed on a model but errored every request — and `n_sessions` is exactly what the board's `$/session` denominator rests on, so the invariant is worth a one-line pin:

```python
# Inside the existing _patch_now(...):
await ledger.record(plan, failed=True, model="m3", session="session-A")
assert K_M_SMONTH.format(
    plan="test-plan", model="m3", period="2025-09") not in redis.hlls
```

The docstring in `record()` says "counted on success only (failure paths pass no session)" — so today this is doubly-protected by the call sites, but the helper's `not failed` clause is the last line of defence and should be tested.

tests/fake_redis.py
```diff
@@ -157,6 +174,23 @@ async def hincrbyfloat(self, key, field, amount):
...
+    # and have bounded error; tests asserting "this many distinct sessions"
+    # want exact answers.
+    async def pfadd(self, key, *members):
+        s = self.hlls.setdefault(key, set())
+        added = 0
+        for m in members:
+            if m not in s:
+                s.add(m)
+                added += 1
+        return added
```
Fledgewing: **[nit] FakeRedis.pfadd return value diverges from real Redis**

Real Redis `PFADD key member [member ...]` returns `1` if at least one new member was added to the HLL, `0` otherwise (independent of how many members were passed). The fake returns the integer count of newly-added members:

```python
async def pfadd(self, key, *members):
    s = self.hlls.setdefault(key, set())
    added = 0
    for m in members:
        if m not in s:
            s.add(m)
            added += 1
    return added
```

No caller in this PR reads PFADD's return value (the HLL cardinality is always read via `pfcount`), so the divergence is invisible today. But the rest of `FakeRedis` is shaped to match real Redis return semantics — `zadd`'s `CH`/`XX` accounting, `incr`'s int, `hincrbyfloat`'s float — and a future caller that branches on `await pipe.pfadd(...) == 1` (the canonical "did I add something" check) would silently read a count > 1 and behave wrong.

Cheap to align now, while the test surface is fresh and the only PFADD caller is one that doesn't care:

```python
return 1 if added else 0
```

Fledgewing:
Reviewed 5 findings: 0 blockers, 1 should-fix, 4 nits — see inline comments at 4ed121e.

Headline: `plan_sessions` is a sum of per-model HLL cardinalities rather than the union, so a session that lands on multiple models within one plan is over-counted in the subscription rate denominator (`monthly_cost / plan_sessions`). Fix without adding a new HLL: queue one `pipe.pfcount(*smonth_keys)` for the union cardinality. The remaining four are nits on the subscription math round-trip cancel, the positional-slice `raw[len(reads):]` style, a missing test for `failed=True + session=`, and `FakeRedis.pfadd`'s return value diverging from real Redis.

switchyard/usage.py
```diff
@@ -186,10 +201,31 @@ async def model_overview(
...
+                plan=plan_key, model=ref, period=month))
+        # Plan-level PFCOUNT: the union of every per-model HLL, returned by
+        # Redis as a single integer. Sums of per-model cardinalities would
+        # double-count any session that touched more than one model of this
+        # plan (e.g. mid-loop spillover between two configured models), and
+        # the subscription rate `monthly_cost / plan_sessions` is sensitive to
+        # that denominator -- an over-counted plan_sessions understates the
+        # board's $/session for every model of a subscription plan.
+        trailing.append(("plan", None))
+        pipe.pfcount(*[K_M_SMONTH.format(
```
Fledgewing, line 222: **[nit] The union cardinality that round 2 fixes for is not directly pinned by any test**

The whole point of this round was `plan_sessions` switching from "sum of per-model HLLs" to "their union", and the fix is the new multi-key PFCOUNT at `pipe.pfcount(*[K_M_SMONTH.format(...) for ref in model_refs])`. But `plan_n_sessions` is never asserted on. A grep for the field across `tests/` returns no hits; the only references are in `switchyard/usage.py` and `switchyard/portal/app.py`.

`test_model_overview_counts_unique_sessions` exercises the per-model PFCOUNTs (one session on m1, none on m2 → `result["m1"]["n_sessions"] == 1`, `result["m2"]["n_sessions"] == 0`) but it doesn't pass the same session to two models, and it doesn't assert on `plan_n_sessions`. So a regression that breaks the union — e.g. someone refactoring `pipe.pfcount(*all_keys)` back to per-model `pipe.pfcount(key)` plus a sum, or `FakeRedis.pfcount(*keys)` losing its `len(keys) == 1` short-circuit — would not be caught by any existing test. The unit tests for `model_effective_cost_per_session` pass `plan_sessions` as an argument, so they exercise the threshold and the rate math but not the union computation itself.

The fix is the focal point of round 2 and benefits from explicit pinning. A four-line test would lock it in:

```python
def test_model_overview_plan_n_sessions_is_union_not_sum():
    """plan_n_sessions = |union of per-model HLLs|, not the sum.

    One session hitting two models of one plan must NOT double-count:
    union cardinality is 1 even though per-model cardinalities both read 1.
    The original implementation summed per-model counts, which over-counted
    and understated the subscription rate -- the round-2 fix is the
    multi-key PFCOUNT here, so the union semantics are the invariant.
    """
    redis = FakeRedis()
    ledger = Ledger(redis)
    plan = fake_plan(metered=True)
    with _patch_now(datetime(2025, 9, 21, 14, 0, 0, tzinfo=timezone.utc)):
        async def record():
            await ledger.record(plan, model="m1", session="shared",
                                prompt_tokens=100, cost=0.01)
            await ledger.record(plan, model="m2", session="shared",
                                prompt_tokens=100, cost=0.01)
            await ledger.record(plan, model="m2", session="unique-m2",
                                prompt_tokens=100, cost=0.01)
        run(record())
        result = run(ledger.model_overview("test-plan", ["m1", "m2"], "2025-09"))
    # Per-model: m1 saw 1 distinct, m2 saw 2 distinct.
    assert result["m1"]["n_sessions"] == 1
    assert result["m2"]["n_sessions"] == 2
    # Plan-level union: the shared session counted once, plus the unique one = 2.
    assert result["m1"]["plan_n_sessions"] == 2
    assert result["m2"]["plan_n_sessions"] == 2
```

The exact numbers (1, 2, 2) are the contract: a regression to sum would report `plan_n_sessions == 3` (1+2) and fail loudly.

Fledgewing:
Reviewed round 2 (HEAD 88d0b27): all five round-1 findings are resolved. The sum-vs-union over-count is fixed by `pipe.pfcount(*smonth_keys)` in `model_overview` and the portal reads `plan_n_sessions` off the row; the subscription round-trip cancel is gone (single division); the positional slice is now decoded against a parallel `trailing` list; the failed-event success-only invariant is pinned by the new `test_record_failed_event_does_not_pfadd`; `FakeRedis.pfadd` returns 1/0 and `FakePipeline`/`FakeRedis.pfcount` accept `*keys` for the union. 34 tests pass, full suite green, render_preview renders.

One new nit posted inline: `plan_n_sessions` — the union cardinality the round-2 fix is specifically for — is never asserted on. Worth a regression test that one session on two models of one plan yields union=1, not sum=2.

Fledgewing:
Round 3 (HEAD 30464e1): clean sign-off. The commit is test-only — `git diff 88d0b27..30464e1` touches `tests/test_usage.py` and nothing else. The new `test_model_overview_plan_n_sessions_is_union_not_sum` pins exactly the invariant the round-2 comment asked for: one shared session on two models + one unique session on the second model → per-model n_sessions (m1=1, m2=2) and plan-level union cardinality (plan_n_sessions=2 on both rows, which would be 3 under the old sum-based regression). Inserted before the runner so it gets collected; 35/35 pass in `tests/test_usage.py`, full `for t in tests/test_*.py` loop green. No new findings.

PR #87, opened 2026-09-22
CLI sidecars: honor max_tokens (truncate + finish_reason "length") or refuse with 400; remove fold_max_tokens (#70)

### LINKED ISSUE #70: CLI sidecars fake-enforce max_tokens as a prompt instruction; models 'say they'll do X' then stop, and finish_reason lies about it — honor max_tokens properly or refuse
## The rule

If the caller passes `max_tokens`, the sidecar honors it (enforce post‑hoc, or refuse up front with 400 if it cannot). If the caller does not, behavior is unchanged — the provider decides output length.

## Symptom

On every CLI-backed plan (claude-max, codex, opencode; claude-go/opencode-go), when a caller passes `max_tokens`, the model frequently answers in the shape *"I'll go check the file and patch the function"* — and then the turn just ends. No file is checked, no patch is made. The agent gets an empty conclusion and either waits forever or hallucinates the missing result. The model was clearly going to do the work; it just stopped mid-plan.

The whole thing is Switchyard's fault, not the model's. The model follows the instruction it was handed. The caller sees the *model* as the one that misbehaved only because nothing in the response says otherwise.

## Where it actually comes from

`sidecars/cli_bridge/server.py:865-879` (`fold_max_tokens`):

```python
def fold_max_tokens(system: str | None, max_tokens: int | None) -> str | None:
    ...
    words = max(10, int(max_tokens * 0.7))
    hint = (f"Answer in at most roughly {words} words. Be direct: no preamble, "
            "no restating the question.")
    return f"{system}\n\n{hint}" if system else hint
```

…called unconditionally on every chat at `sidecars/cli_bridge/server.py:1482`:

```python
system = fold_max_tokens(system, body.get("max_tokens"))
```

So a caller's `max_tokens: 400` arrives at the CLI as a system instruction that *reads* like an enforcement (*"be terse, no preamble"*) but isn't one. On a model that was about to do real work, the "be terse" part gets obeyed first — the model cuts the narration short, including any "I'll go check …" prefix. Work is not done; quota is paid for whatever the model did emit.

The second half is what makes the bug invisible. `to_openai` at `sidecars/cli_bridge/server.py:1188-1235` hardcodes the finish reason:

```python
"choices": [{
    "index": 0,
    "message": {"role": "assistant", "content": payload.get("result", "")},
    "finish_reason": "stop",
}],
```

A response that the model itself cut short by following our instruction looks identical on the wire to a clean stop: `finish_reason: "stop"`, `content: <whatever the model said>`. OpenCode/Codex/Claude Code all key tool-loop continuation off this field, so a `"stop"` frame after *"I'll go check …"* sends the next turn on assuming the work was already done. The caller has no way to say *"don't trust this one, ask again."*

`mcp_bridge` doesn't escape this — non-tool requests fall through to `cli_bridge._handle_chat` at `sidecars/mcp_bridge/server.py:1415`, so the same fold lands on every MCP-backed plan that doesn't have native tools.

## Why this is the bug, not a tradeoff to preserve

The current behaviour exists because no CLI accepts a real token cap (Claude Code offers `--max-turns`, Codex has no flag at all), and *some* reduction felt better than none (commit `5429a64`, README:761, `/health` reports `enforces_max_tokens: false`). The intent was reasonable, but a "soft" pad that goes in the system prompt has three unforgivable properties:

1. It is not enforcement. A model can and does exceed the suggested word count; what it must not do is narrate work it then doesn't perform, and that is exactly what is happening.
2. The stall-out is silent. `finish_reason: "stop"` is a lie the moment the model obeyed our "be terse" instruction by cutting its own answer short. The caller cannot detect.
3. There is no valid "default" mode where `max_tokens` is a soft hint. The field means what it says; either honor it or refuse the request.

So fold-as-instruction is removed. There is no per-plan toggle to keep it.

## The right shape: honor `max_tokens`, always

The contract a caller can rely on is one of two outcomes — neither of which is "the field was quietly suggested." Pick whichever the sidecar can deliver on its lane.

**1. Enforce post-hoc and tell the truth.** The CLI has no token cap, but we own what comes back. After the CLI finishes, count tokens with a tiktoken/ttrc-bytes estimate. If we're over the caller's `max_tokens`, truncate the `content` to fit and emit `finish_reason: "length"`, never `"stop"`. Same shape on the SSE terminal frame. The CLI's lack of a flag is irrelevant once we own the bytes; the caller's cap is honored exactly.

**2. Refuse honestly when we can't.** If the sidecar cannot enforce (e.g. a streaming path that flushes before we can recover the full output, or a plan with truly unbounded output), return `400` with `"max_tokens is not enforceable on this lane"`. The caller drops the field or routes to an API-keyed lane that does honor it. Better than silently not enforcing and lying in `finish_reason`.

Mode 1 wherever the buffering cost is acceptable (the existing CLI bridge already buffers the full output before framing it). Mode 2 for streaming and for plans that genuinely cannot enforce (codex today). Same contract on every lane: **no quiet violation of `max_tokens`.**

The fold-to-instruction shortcut is removed entirely. So is `enforces_max_tokens: false` from `/health` — that field was advertising the bug as a feature; the health field now reports the *mode* (`enforce` | `refuse` with reason) so a caller can see what it's getting.

## Why this isn't a per-plan toggle

A toggle that lets an operator keep the buggy fold mode preserves a bug as a setting. The right shape is one enforcement semantics across the bridge — fold-as-instruction stops being an option. Plans that genuinely cannot enforce (streaming, codex) refuse with 400; everything else honors the cap. There is no third "warn the caller and ignore" path.

## Files touched

- `sidecars/cli_bridge/server.py:865-879` — `fold_max_tokens` removed (not deprecated; folded into the wall). Call sites no longer exist.
- `sidecars/cli_bridge/server.py:1482` — replaced by post-hoc counting + truncation; `finish_reason: "length"` when the cap is hit.
- `sidecars/cli_bridge/server.py:1188-1235` (`to_openai`) — emit `"length"` for truncated replies, never `"stop"`. Streaming: same field on the terminal SSE frame; pre-flight refusal uses 400.
- `sidecars/mcp_bridge/server.py:1415` — non-tool path picks up the same enforcement.
- `/health` (`enforces_max_tokens`) — reports mode + reason instead of `false`. A plan that *can* enforce reports `true`; a plan that refuses reports `false` with `"streaming-only"` or `"no-truncation-flag"` in `enforces_max_tokens_reason`.
- `README.md:761` — paragraph rewritten: these lanes honor `max_tokens` (truncate + `length`) or refuse with 400 if they cannot.
- `tests/test_cli_bridge.py:237` — `test_max_tokens_becomes_an_instruction` deleted; replaced with `test_max_tokens_truncated_with_length_reason` and `test_unenforceable_max_tokens_returns_400`.

## Acceptance criteria

- [ ] Caller sends `max_tokens: N` and N is reached: sidecar truncates and emits `finish_reason: "length"` (not `"stop"`).
- [ ] Caller sends `max_tokens` on a plan that cannot enforce: sidecar returns 400 `"max_tokens is not enforceable on this lane"` before any upstream call is made.
- [ ] Caller omits `max_tokens`: behavior unchanged (the provider decides output length).
- [ ] No code path uses prompt-instruction padding as an enforcement mechanism. `fold_max_tokens` and its callers are gone.
- [ ] Streaming replies honor the same semantics: a truncated stream ends with `finish_reason: "length"`; a pre-flight refusal can still fire 400 before the SSE header.
- [ ] `/health` reports `enforces_max_tokens: true` for plans that enforce and `false` (with reason) for plans that refuse.
- [ ] No regression on API-keyed lanes — they already enforce; the change is a no-op there.

Fledgewing:
## Why
Callers passing `max_tokens` on CLI-backed plans got a system-prompt *suggestion* ("answer in ~N words, be terse") instead of enforcement: models obeyed the terseness by answering "I'll go check X" and stopping, and `finish_reason: "stop"` looked identical to a clean exit. See #70.

## What
- Removed `fold_max_tokens` and its unconditional call — no prompt-instruction fake enforcement remains.
- Added `estimate_tokens` (UTF-8 byte/4 estimate, dependency-free) and `enforce_max_tokens(payload, max_tokens)`: content truncated to the cap, `finish_reason: "length"` in both the JSON choice and the SSE terminal frame; `usage` still reports what the CLI actually billed.
- `_handle_chat` probes the profile: claude/opencode lanes enforce post-hoc; codex lanes refuse up front with `400 "max_tokens is not enforceable on this lane"` (reason `no-truncation-flag`) before any spawn/gate/SSE header, unless `max_tokens` is omitted.
- `/health` now reports `enforces_max_tokens: true/false` plus `enforces_max_tokens_reason` when false (cli_bridge and mcp_bridge).
- Replaced `test_max_tokens_becomes_an_instruction` with truncation/400/health/`fold_max_tokens is gone` tests (48→52 cli-bridge, 55→56 mcp-bridge). README paragraph and TESTING.md smoke note updated.

## Semantics
- `max_tokens` absent or <=0: byte-for-byte unchanged behavior.
- Enforcing lane over cap: truncate + `"length"`.
- Non-enforcing lane (codex): 400 before any upstream call.
- API-keyed lanes: untouched (already litellm-native enforcement).

## Verification
`python3 tests/test_cli_bridge.py`, `python3 tests/test_mcp_bridge.py`, offline suite loop, `python3 tests/render_preview.py` — all green (suite-wide pre-existing env-dep FAILs unchanged). 52 cli-bridge tests passed, 56 mcp-bridge tests passed.

Fixes #70


sidecars/cli_bridge/server.py
```diff
@@ -1632,6 +1676,23 @@ async def _handle_chat(body: dict):
                 "type": "tools_unsupported"}})
 
     messages = body.get("messages") or []
+
+    # Refuse max_tokens on lanes that cannot honor it honestly (issue #70).
+    # This is BEFORE the tools-refusal block: a request carrying both max_tokens
+    # AND tools would otherwise get a tools-unsupported 400 instead of the
+    # more specific max_tokens reason. The placement also guarantees the 400
+    # is raised before any stage_or_fail / flatten / gate / spawn, so no SSE
+    # header is ever sent for a request we already know to refuse.
```
Fledgewing: **[should-fix] Comment claims max_tokens block fires BEFORE the tools-refusal block, but the code places it AFTER.**

In `sidecars/cli_bridge/server.py` `_handle_chat`, the `tools_unsupported` 400 is at line 1669 (existing) and the new `max_tokens_unenforceable` block is at line 1686 (after `messages = body.get(...)`). The diff context confirms this: the inserted block sits in the `+` lines immediately following the `messages = …` assignment, well below the tools check that already raises.

The inline comment at line 1680-1685 claims:

> This is BEFORE the tools-refusal block: a request carrying both max_tokens AND tools would otherwise get a tools-unsupported 400 instead of the more specific max_tokens reason.

That contract is not what the code now does. A codex request that carries both `tools` and `max_tokens` on cli_bridge hits the tools check first and gets `type: "tools_unsupported"`, never reaching the max_tokens refusal. (mcp_bridge routes tool requests to `handle_tool_request` so the same request never reaches `_handle_chat` from there — this only manifests for direct cli_bridge callers, which is the case the comment is reasoning about.)

Suggested fix: either move the `max_tokens_unenforceable` block above the `tools_unsupported` block so the code matches the comment's stated intent, or rewrite the comment to match the actual ordering (e.g., "after the tools-refusal block: tools takes priority because both 400s are pre-flight and tools is the more general refusal for a CLI-backed plan").

This is not a contract violation against issue #70's acceptance criteria — both 400s fire before any upstream call/SSE — but the comment is asserting a specific ordering the code does not implement, and a future maintainer chasing "the comment is wrong" will be left guessing which side to move.

Fledgewing:
Reviewed PR #87 at cf32d50: contract points from issue #70 hold (omitted/<=0 max_tokens pass-through with `finish_reason: "stop"`; enforcing lanes truncate + `"length"` with `usage` left untouched; codex returns 400 `max_tokens_unenforceable` with the exact message string before any upstream call/SSE header; `to_openai` default `finish_reason="stop"` preserved on mcp `render_turn "final"`; `sse_from_completion` delta→terminal→[DONE] order intact; tests inserted before the `__main__` runners, no real provider calls; mcp `/health` mirrors cli_bridge profile). One finding posted — see inline comment.

Fledgewing:
Round-2 re-review at aa1cb1c: round-1 finding fully addressed. The `max_tokens_unenforceable` block was moved from below the `tools_unsupported` check to above it (now at `sidecars/cli_bridge/server.py:1666`), with the comment wording updated to "Placed BEFORE the tools_refusal block" so comment and code now match; `test_unenforceable_max_tokens_returns_400` gained an ordering assertion that codex+tools+max_tokens now returns `max_tokens_unenforceable` rather than `tools_unsupported` (the test fires and the suite reports 52 cli-bridge / 56 mcp-bridge green). Only those two files were touched in this commit (`git diff cf32d50..aa1cb1c --stat`), the move is a pure relocation with no logic changes, and tracing the remaining scenarios (codex + tools only -> tools_unsupported; claude/opencode + max_tokens -> post-hoc truncate; mcp_bridge tool path still routed to `handle_tool_request` and ignores max_tokens) shows no behaviour drift. No new findings.

PR #90, opened 2026-09-23
Fix issue #86 ("apply.sh recreates / restarts containers in place"): replace apply

### LINKED ISSUE #86: apply.sh recreates / restarts containers in place — no zero-downtime path for gateway, portal, or sidecars
## Symptom

`scripts/apply.sh` is documented as "the one command" for rolling a config or code change out to the live stack, but every code path through it takes the gateway, portal, and (on env-key adds) every sidecar offline simultaneously:

1. **`apply.sh:175-176`** — when `scripts/sync-env.sh` appended any new `.env` key, `docker compose up -d --force-recreate gateway portal claude-max-sidecar codex-sidecar opencode-go-sidecar xai-token-proxy` stops all six, then brings them up sequentially. The gateway's `start_period: 40s` + `interval: 15s` × `retries: 5` healthcheck means the port on `:4000` can stay unreachable for ~75–115s while the new container warms. Every CLI tool pointed at the gateway (Claude Code, Codex, opencode, anything using `/v1/messages` or `/v1/chat/completions`) gets `connection refused` / `ECONNRESET` for the duration.

2. **`apply.sh:179`** — when no env keys were added, plain `docker compose up -d` still recreates any service whose **definition** changed: rebuilt image, added env, new mount. A code change to `switchyard/*.py` triggers a rebuild (`apply.sh:159`) and then this `up -d` recreates the gateway in place. Same ~75–115s of gateway downtime.

3. **`reload.sh:93-94`** — when the router signature differs (new model string, new `api_base`, new credential env), `docker compose restart gateway` is unconditional. The restart sequence is `docker stop` (10s grace, then SIGKILL) → `docker start` → `start_period` → healthchecks → ready. The `start_period: 40s` is the dominant term here too. Any LiteLLM worker with an in-flight streaming response is killed mid-stream, and the caller's tool sees `httpx.RemoteProtocolError` mid-tool-loop.

4. **Sidecar code / image changes** — `apply.sh:159` rebuilds the `sidecar` image used by `claude-max-sidecar`, `codex-sidecar`, `opencode-go-sidecar`, `opencode-go2-sidecar` together, and the subsequent `up -d` (or `--force-recreate`) drops each one. The gateway has no internal retry for a sidecar socket failure (`hooks.py` uses `num_retries=0`, see the comment at `hooks.py:11-18`); the caller's 429 retry re-enters the proxy and re-picks against the fresh cooldowns, but a *connection refused* during recreate surfaces as 502 to the caller, not as a re-pickable failure.

The hot-swap path in `reload.sh:67-87` is the only zero-downtime leg, and it only fires when `models.router_signature` is byte-equal — i.e. policy-only edits. Any router-shaped edit, any code change, any new env key: gateway goes down.

This is a **bug**, not an enhancement, because the existing behaviour is unsafe to run during a working session. `#15` was the last symptom of this — "242 'socket connection was closed' in one day" — and closed because the immediate cause (a specific socket-drop path) was patched. The structural cause (the deploy itself drops the gateway) is unchanged.

## Repro

On the main checkout, with the stack running:

```bash
# Edit config/plans.yaml to add a new model to an existing lane.
# This is a router-shaped change: signature differs, plan cap may change.
scripts/apply.sh
# Watch the gateway port:
watch -n 1 'docker compose ps gateway; \
  curl -sS -o /dev/null -w "%{http_code} %{time_total}s\n" \
    -m 1 http://localhost:4000/health/liveliness || echo down'
```

Expected after the fix: gateway stays `200` throughout the whole apply. Observed today: `connection refused` / `000` for the full `start_period` (40s+) plus any extra healthcheck retries.

Force-recreate path (any new `.env` key):

```bash
echo "NEW_KEY=1" >> .env.example
scripts/sync-env.sh                  # appends NEW_KEY to .env
scripts/apply.sh                     # --force-recreate on all six
# Observed: gateway, portal, and all four sidecars go down in turn;
# the gateway is the last to come back up (~75-115s).
```

## Where the downtime comes from

```
apply.sh:159   docker compose build "${stale[@]}"             # builds images; running containers unaffected — OK
apply.sh:175   docker compose up -d --force-recreate \
                    gateway portal claude-max-sidecar codex-sidecar \
                    opencode-go-sidecar xai-token-proxy       # ALL six recreated, stop-then-start
apply.sh:179   docker compose up -d                            # recreates services whose definition changed
                 (gateway env, gateway mounts, rebuilt gateway image, sidecar rebuild, ...)
reload.sh:94   docker compose restart gateway                  # unconditional when router sig differs
```

Three recreate sites, all in-place. No service is run twice. No port is held by a stand-by. The `depends_on` block on the gateway (`docker-compose.yml:73-74`) means compose waits for `redis`/`postgres` to be `service_healthy` before starting the new gateway, but it does **not** wait for the **old** gateway to drain — `up -d` (and `--force-recreate`) defaults to `stop-then-start` with the standard 10s SIGTERM grace, then SIGKILL.

The `healthcheck` on the gateway (`docker-compose.yml:75-80`) is good for "did the new container come up" but not for "can I take this one out". There is no `pre_stop` hook, no readiness gate on the **outgoing** container, and no overlap between the two.

## Refined approach: drain flag + session migration, in unload-first order

For the sidecars, true zero-downtime per sidecar is hard (each owns one CLI login, so two instances means two logins — that's exactly what `opencode-go2-sidecar` exists for as separate-quota, not HA). What *is* feasible — and is enough to keep the LANE up — is **drain + migrate + replace**, ordered so the lane's absorbing sibling is up and ready before it has to take load.

### Picker changes (one new gate, one new procedure)

1. **New Redis key** `K_DRAIN = "sy:drain:{plan}"` in `slots.py`. Written by `apply.sh`, read by the picker.

3. **Picker honors the drain flag** in `picker.py:_visit_ref` (`picker.py:225-281`), alongside the existing `cooldown_state` check at line 244. A drained plan is skipped for **new** requests; the body walk falls through to the lane's next sibling plan. Skipped reason: `f"{ref}(draining)"`.

3. **Session migration** — the picker honors session leases via `get_lease` (`picker.py:643-686`), and an unmodified lease points the next turn straight back at the drained sidecar. `apply.sh` has to rewrite the lease before stopping the container:

   - Maintain a reverse index `K_LEASE_PLAN = "sy:lease_plan:{plan}"` as a Redis set, written in `slots.set_lease` (`slots.py:249-250`) alongside the existing lease write, removed in `slots.drop_lease` (`slots.py:255-261`) alongside the existing delete. SCAN-based enumeration of `sy:lease:*` is too slow at scale; this set is O(N) per drain where N is the sidecar's active sessions.
   - For each session ID in the set, re-pick using the picker's normal routing logic with the drained plan excluded. That gives a sibling `model.ref`, and the lease is rewritten: `set_lease(session, new_ref, ttl)`. The picker itself doesn't need to know migration happened; it just sees the new lease on the next request.
   - **Known cost of mid-session migration** — CLI bridges (`mcp_bridge`, `cli_bridge`) keep conversation history in the gateway's request payload, so the new CLI process continues the conversation fine. CLI-side session resumption (e.g. `claude --resume`) loses its in-memory scratch state, but only for the CLI process itself; the gateway-side turn-to-turn continuity is unaffected. This is the same shape as a normal cooldown re-route, just operator-triggered.

4. **In-flight wait** — before stopping the sidecar, `apply.sh` polls `ZCARD sy:inflight:{plan}` until it hits 0 (existing `slots.in_flight(plan)` already prunes stale entries: `slots.py:192-195`). Two safety nets cover what polling can't:
   - A hard grace period (`apply.sh --drain-grace-secs`, default 30s) past `ZCARD == 0`, so any tool call that just claimed a slot has time to settle.
   - For `mcp_bridge` parked calls, the grace has to clear the parked-call timeout (90s); make the grace configurable per-sidecar or just default to 120s for the MCP-bridged ones. The picker doesn't see parked calls; the gate would have to be a separate `sy:parked:{plan}` counter, or the grace is the simpler answer.

### Per-sidecar drain order — unloaded first

A single sidecar's load right now is `ZCARD sy:inflight:{plan}` plus a recent-activity rate from the usage ledger. Order the sidecars being replaced like this:

```
score(plan) = in_flight + recent_rate
sort by score ascending        # 0 → light → heavy
```

…then walk them in that order. The reasoning, lane-by-lane:

- **Multi-sidecar lane** (e.g. two `opencode-go` plans sharing one lane, or any lane with sibling plans): when the unloaded sibling is drained first, the lane's other members take the brief load spike from migrated sessions *while still healthy*. When the heavier sibling is drained next, the first sibling is **already back up and warm**, ready to absorb. Replacing heavy first would have the absorbing sibling carrying its own load PLUS the migrating sessions PLUS still warming up — three multiplicative sources of degradation.
- **Single-sidecar lane** (most lanes today — `claude-max`, `openai`+`astra` on one `codex-sidecar`, `grok` on `xai-token-proxy`): draining always takes the lane down briefly because there's no sibling to absorb. The order doesn't help availability, but it does help **operator feedback**: an unload-first walk surfaces a broken image or auth failure on the cheapest replacement first, before doing the more disruptive work on the loaded ones. By the time the heavy walk starts, the easy ones have either succeeded (good signal) or failed loudly (saved time).
- **`xai-token-proxy`** is the odd one out — no CLI process, just an OAuth token in a file. Replacement is trivially cheap regardless of load. Still goes through the same drain flow because the gateway calls it over HTTP the same way as the CLI sidecars; the cost is just lower.

### `apply.sh` shape

Replace the single `docker compose up -d [--force-recreate]` line at `apply.sh:175` / `apply.sh:179` with a per-sidecar drain loop. The gateway and portal still get the same proxy-based treatment from the previous revision of this issue.

```bash
# After build, before up -d:
SERVICES=$(docker compose config --services | grep -E 'sidecar|token-proxy')
for svc in $(order_by_load "$SERVICES"); do
  plan=$(docker compose exec -T "$svc" printenv SWITCHYARD_PLAN)
  redis-cli -n 1 SET "sy:drain:$plan" 1                   # picker skips this plan
  migrate_sessions "$plan"                                # rewrite leases to siblings
  wait_inflight_drain "$plan" --grace "${GRACE:-120}"
  docker compose stop "$svc"                              # SIGTERM, in-flight should be 0
  docker compose up -d --no-deps --force-recreate "$svc"  # new image, new container
  wait_healthy "$svc"
  redis-cli -n 1 DEL "sy:drain:$plan"
  settle "$svc"                                           # brief, so the next drain starts from a clean lane
done

# Then the gateway/portal proxy-aware swap from the prior revision.
# Then `docker compose up -d` for any service not handled by the above.
```

The `--dry-run` path prints the planned drain order without writing any state. The `--build` flag still triggers `docker compose build` first, so a rebuild is paired with the per-service recreate; without `--build`, the drain loop is a no-op when no image is stale.

### What this does and doesn't get us

- **Lane stays up when a sibling exists.** For the multi-sidecar case, the lane is observably 200 throughout the swap. The user's active turn on the doomed sidecar finishes; their next turn is on the sibling.
- **Lane goes down briefly when no sibling exists.** For the rest, the lane takes sub-second-to-30s of downtime per sidecar, ordered so the easy ones go first. Better than today's 30s+ of `connection refused` for every sidecar, but not zero.
- **Gateway + portal still need the proxy-based fix.** The drain+migrate path solves sidecars. The gateway and portal themselves (`:4000`, `:4001`) need the two-instance rolling swap described in the prior revision.
- **CLI auth state is lost per sidecar.** The CLI process inside the new container starts fresh. Sessions that were mid-tool-loop on the old CLI process get their tool result re-sent (the gateway replays from the conversation history), and the new CLI process resumes from the gateway's view. This is exactly how a cooldown re-route behaves today; the only new thing is that the operator triggered it.

### Acceptance

After the fix:

- `scripts/apply.sh` while a Claude Code session is mid-tool-loop on `claude-max`: the tool call finishes on the old sidecar; the next user turn lands on a new sidecar without the user seeing `connection refused` or a stalled UI.
- `scripts/apply.sh` while a session is hot on `opencode-go`: the session migrates to `opencode-go2` (or the lane's other sibling) before the sidecar is replaced, and the lane stays at 200 throughout.
- `scripts/apply.sh --dry-run` shows the drain order (sorted by current load), the number of leases to migrate per sidecar, and the predicted total drain time.
- The repro at the top of this issue shows `200` for the gateway port throughout, and at most a sub-second blip per sidecar on the lane's `docker compose ps` healthcheck.

Fledgewing:
## Summary

Fix issue #86 ("apply.sh recreates / restarts containers in place"): replace
apply.sh's stop-then-start recreation of sidecars, xai-token-proxy, gateway
and portal with a zero-downtime, unload-first drain loop. The picker gains
a Redis-backed drain flag (`sy:drain:{plan}`) that skips a plan for new
requests; apply.sh sets the flag, migrates the drained plan's session leases
to lane siblings via `switchyard.drain`, waits for in-flight slots to hit
zero (plus a `--drain-grace-secs` grace window that covers mcp_bridge's 90s
parked-call timeout), then stop → recreate → wait-healthy → clear flag, per
sidecar, ordered lightest-first (ascending `ZCARD sy:inflight:{plan}`).
Gateway and portal are recreated fold-on-at-a-time after the sidecar loop
(still a brief blip, but isolated); reload.sh's policy-only hot-swap path
is preserved.

## Workstream 1 — picker drain gate + lease-plan reverse index + session migration helper (python)

Files: `switchyard/slots.py`, `switchyard/picker.py`, `switchyard/drain.py`, `tests/test_drain.py`

- `slots.py`: new keys `K_DRAIN = "sy:drain:{plan}"` and reverse lease
  index `K_LEASE_PLAN = "sy:lease_plan:{plan}"` (a Redis SET of session
  ids, TTL matched to the lease). Maintained by `set_lease`, `touch_lease`,
  and `drop_lease` so a stale SET is impossible.
- `slots.py`: `set_lease(session, ref, ttl, plan_key=None)` — ref is the
  model ref (`plan/model`); `plan_key` is the prefix, derived if not
  passed (so existing call sites stay compatible).
- `slots.py`: new methods `set_drain(plan)`, `clear_drain(plan)`,
  `is_draining(plan)`, `sessions_on_plan(plan)`.
- `picker.py::_visit_ref`: drain gate next to the cooldown gate. A drained
  plan surfaces as `(draining)` in `considered` and falls through to the
  lane's next sibling. The affinity path does NOT need a drain check —
  apply.sh migrates leases before draining, so no lease points at the
  drained ref during its window.
- `picker.py`: `set_lease` now passes the plan key alongside the ref.
- `switchyard/drain.py` (new): `migrate(plan_key, slots, picker, ttl=None)`
  walks `sessions_on_plan(plan_key)`, calls `picker.pick(lane, session,
  exclude=drained_refs)` per session, rewrites the lease via
  `set_lease(session, new_ref, ttl, new_plan.key)`, prints one line per
  migrated session and a count. CLI entry point:
  `python -m switchyard.drain <plan> [--ttl N]`.
- `tests/test_drain.py` (new, 4 tests): a drained plan surfaces as
  `(draining)` and the sibling absorbs; clearing the flag is a sentinel-
  only operation (no state leakage); set/touch/drop keep the reverse index
  in sync; `drain.migrate` rewrites N leases to a sibling and leaves the
  non-drained plan's leases untouched.

## Workstream 2 — apply.sh: unload-first per-sidecar drain loop (shell)

Files: `scripts/apply.sh`, `TESTING.md`

- `apply.sh`: new flag `--drain-grace-secs N` (default 120; covers
  mcp_bridge's 90s parked-call timeout).
- `apply.sh`: rewrite of step 4 — per sidecar, in unload-first order
  (ascending `ZCARD sy:inflight:{plan}`, computed on the host via
  `docker compose exec -T redis redis-cli -n 1`; service list = `docker
  compose config --services | grep -E 'sidecar|xai-token-proxy'`; plan via
  `docker compose exec -T \$svc printenv SWITCHYARD_PLAN`):
  1. `SET sy:drain:{plan} 1`
  2. `docker compose exec -T gateway python -m switchyard.drain \"\\"` (depends on WS1; failures are logged, not fatal)
  3. Poll `ZCARD sy:inflight:{plan}` until 0 or the grace window, then honour the full grace (so a parked MCP call can land)
  4. `docker compose stop \$svc` → `docker compose up -d --no-deps --force-recreate \$svc`
  5. Wait healthcheck green (`docker inspect\ .State.Health.Status`)
  6. `DEL sy:drain:{plan}`
  7. Short settle sleep before the next sidecar
- `apply.sh`: portal and gateway are recreated fold-on-at-a-time AFTER
  the sidecar loop (portal first, then gateway), never all at once.
- `apply.sh`: a service not owning a stale image (and no `env_added`)
  is skipped from the drain loop — config-only edits are still a no-op
  (the mounted plans.yaml is re-read on next request).
- `apply.sh`: `--dry-run` prints the planned drain order with per-
  sidecar `in_flight`, `leases`, and a predicted total drain time.
- `apply.sh`: when `reload.sh` takes the policy-only hot-swap path,
  apply.sh does NOT additionally force-recreate the gateway.
- `TESTING.md`: new section 11 with the live zero-downtime apply.sh
  drain runbook (11a repro, 11b success criteria, 11c dry-run check,
  11d grace window tuning).

## Verification

```
python3 tests/test_drain.py        # 4/4
python3 tests/test_routing.py      # all routing tests passed
python3 tests/test_hot_reload.py   # all hot-reload tests passed
for t in tests/test_*.py; do python3 "\" >/dev/null || echo "FAIL \"; done
                                  # 19/19 PASS
python3 tests/render_preview.py    # rendered index.html + fragments
```

Closes #86


switchyard/picker.py
```diff
@@ -245,6 +245,14 @@ async def _visit_ref(self, ref: str, ctx: _VisitCtx,
         if cooled:
             ctx.skipped.append(f"{ref}(cooled: {reason})")
             return None
+        # Drain gate: a plan with the drain flag set refuses NEW requests.
+        # apply.sh migrates existing leases before flipping the flag, so the
+        # affinity path can ignore this — the only way a NEW request reaches a
```
Fledgewing: **[blocker] Drain-gate comment claims an ordering apply.sh does not honour; the picker’s actual safety comes from a different path**

The comment at lines 248-252 says "apply.sh migrates existing leases before flipping the flag, so the affinity path can ignore this". Look at `scripts/apply.sh`: line 375 does `SET sy:drain:{plan} 1` and line 384 only then runs `python -m switchyard.drain`. The flag is flipped **before** migration, the opposite of the comment’s premise. The same false claim is repeated in the `K_DRAIN` block in `switchyard/slots.py:23-24`.

The picker’s resilience in spite of that ordering is incidental: `Picker._affinity` at line 658-660 calls `drop_lease` whenever `held in ctx.exclude`, so a migration that calls `picker.pick(lane, session, exclude=drained_refs)` drops the lease on the drained plan itself via the affinity drop, not via any apply.sh ordering. The next request from that session has no lease, walks the body, and the body-walk drain gate at line 253 picks the sibling.

That is fragile. A future maintainer who reads "migrate before flag" and either (a) reorders apply.sh to match the comment, or (b) tightens `_affinity` to stop dropping leases it considers stale, will break the load-bearing "affinity drop on held-in-exclude" safety. There is nothing in the current text that warns them.

Smallest durable fix: add a defensive `is_draining(plan.key)` check at the top of `_affinity` (next to the `held in ctx.exclude` block), so the picker is safe regardless of apply.sh’s order. Then rewrite the comments at picker.py:248-252 and slots.py:23-24 to describe the actual contract (the picker refuses to honour a lease on any plan that is currently draining, period).

scripts/apply.sh
```diff
@@ -164,19 +191,298 @@ else
...
+    # (2) hand off in-flight sessions to lane siblings. WORKSTREAM 1 must
+    # have landed for this command to exist; if it hasn't, the picker still
+    # respects the drain flag (new requests skip the plan), so the worst
+    # case is leases pointing at the drained ref until they expire
+    # naturally — logged, not fatal.
+    drain_out="$(mktemp -t switchyard-drain.XXXXXX.txt 2>/dev/null || true)"
+    if docker compose exec -T gateway python -m switchyard.drain "$plan" \
+         >"$drain_out" 2>&1; then
+      migrated="$(tr -d '\r\n' < "$drain_out" | tail -n 1 \
+                    | grep -oE '[0-9]+' || echo 0)"
```
Fledgewing: **[should-fix] Drain-count parsing extracts every digit run, not the count**

`drain.migrate`’s stdout is per-session lines (`{session}: {plan}/{label} -> {ref}`) plus the final `migrated <count> session(s) off <plan>`. Every field there can carry digits: session ids are user-supplied, plans have labels with digits, refs end in `/m3`-style model names. The pipeline here is

```bash
migrated="$(tr -d '\r\n' < "$drain_out" | tail -n 1 \
                | grep -oE '[0-9]+' || echo 0)"
```

`tr -d '\r\n'` collapses every line into one, `tail -n 1` keeps the whole thing, and `grep -oE '[0-9]+'` returns **one match per digit run** — order preserved. Trace the test fixture alone: three sessions named `s-drain-1/2/3` migrated onto `minimax-max/m3` produce `s-drain-3: .../m3 s-drain-2: .../m3 s-drain-1: .../m3 migrated 3 session(s) off minimax-ultra`. `grep -oE '[0-9]+'` returns `3 3 3 2 3 1 3 3` (the count happens to be the last 3); `$(...)` flattens newlines to spaces and `${migrated:-0}` then prints

```
    sessions migrated: 3 3 3 2 3 1 3 3
```

— wrong by every count past the first, AND wrong on the count for any future plan whose name carries digits (e.g. a `5g` plan: `5 3 5g`).

Parse only the count. Two durable options:

```bash
# 1. Parsing-only the count on the migrate line
migrated="$(awk '/^migrated / {print $2; exit}' "$drain_out")"
[ -n "$migrated" ] || migrated=0
```

…or have `drain.py` print the count on its own dedicated line:

```python
sys.stdout.flush()
print(f"count={count}")
```

…and parse `awk -F= '/^count=/ {print $2}'`. Either keeps the per-session audit lines intact for the operator.

switchyard/slots.py
```diff
@@ -246,19 +259,70 @@ async def get_lease(self, session: str) -> str | None:
...
+        await self.redis.sadd(K_LEASE_PLAN.format(plan=plan_key), session)
+        await self.redis.expire(K_LEASE_PLAN.format(plan=plan_key), ttl)
 
     async def touch_lease(self, session: str, ttl: int) -> None:
         await self.redis.expire(K_LEASE.format(session=session), ttl)
+        # The reverse index follows the lease TTL too — if the lease is
+        # still alive, the session is still a candidate for the drain flow.
+        # Reading the lease is cheaper than threading the plan through every
+        # touch_lease call site, and the ref's plan key is on it.
+        lease = await self.redis.get(K_LEASE.format(session=session))
```
Fledgewing: **[should-fix] `touch_lease` GETs the lease after EXPIREing it; the SET TTL refresh silently skips on a deleted lease**

```python
async def touch_lease(self, session: str, ttl: int) -> None:
    await self.redis.expire(K_LEASE.format(session=session), ttl)
    # The reverse index follows the lease TTL too ...
    lease = await self.redis.get(K_LEASE.format(session=session))
    if lease:
        ref = lease.decode() if isinstance(lease, bytes) else lease
        plan_key = ref.split("/", 1)[0]
        await self.redis.expire(K_LEASE_PLAN.format(plan=plan_key), ttl)
```

The `EXPIRE` and the later `GET` are two round-trips. If the lease has been deleted between them — TTL fired, a concurrent `drop_lease` ran, an operator `DEL`d it, the failover promoted a stale primary that doesn’t have the key — `redis.get` returns `None` and the `if lease:` branch silently does nothing. The SET’s TTL keeps whatever it had at `set_lease` time, which by definition is the *old* TTL: this very method is supposed to refresh it. After enough heartbeats-on-a-just-expired-lease, the SET ages out while the lease is being continually recreated (some other code path touched the SET earlier), and `sessions_on_plan(plan)` under-reports.

Downstream consequence: `drain.migrate` walks `sessions_on_plan(drained_plan)` and skips sessions whose entry has been SREMmed by the SET’s natural expiry but whose lease is otherwise alive. Those sessions continue to honour `held` via affinity in `_affinity`. Today they happen to land on a drained ref only when `pick` is called with `exclude=drained_refs` (the affinity drop catches it), but on a normal request they keep using a soon-to-die ref.

Durable fix in the same shape as the rest of the file: a tiny `register_script` keyed by the lease key alone, that does EXPIRE-on-lease, then conditional-EXPIRE-on-SET keyed by the parsed plan prefix, atomically. Lua:

```lua
-- KEYS[1] = lease key
-- ARGV[1] = ttl
local v = redis.call('GET', KEYS[1])
if not v then return 0 end
redis.call('EXPIRE', KEYS[1], ARGV[1])
local plan = string.match(v, '^([^/]+)/')
if plan then
  redis.call('EXPIRE', 'sy:lease_plan:' .. plan, ARGV[1])
end
return 1
```

…and `drop_lease` uses the same shape (EXPIRE-or-DEL + SREM, in one script). The file already shows the pattern (`_CLAIM`, `_BUMP_STREAK`, `_BUMP_AND_COOL`), so this is on-style.

switchyard/drain.py
```diff
@@ -0,0 +1,110 @@
...
+    migrated: list[str] = []
+    for session in sessions:
+        placed = None
+        for lane in relevant_lanes:
+            try:
+                placed = await picker.pick(
+                    lane, session, exclude=drained_refs)
+                break
+            except Exception as exc:
+                last_exc = exc
```
Fledgewing: **[nit] `last_exc = exc` is assigned but never read**

The `except Exception as exc:` branch sets `last_exc = exc` only for the local loop variable to be discarded at the end of the session loop. Either consume it — surface the last placement failure alongside the SKIP line, so an operator chasing a clean “migrated 0 sessions” can see *why*:

```python
if placed is None:
    detail = repr(last_exc) if last_exc else "no live member"
    print(f"{session}: SKIP ({detail})", file=sys.stderr)
    continue
```

…or drop the assignment. Today the code reads like an except block that wanted to do something with the failure and forgot, which makes a future grep for `last_exc` look intentional when it isn’t.

switchyard/drain.py
```diff
@@ -0,0 +1,110 @@
...
+    Returns (count_migrated, list_of_session_ids_migrated). The list is in
+    the order sessions were processed (which is the SET iteration order).
+    """
+    registry = picker.registry
+    lease_ttl = ttl if ttl is not None else registry.settings.lease_ttl_seconds
+    drained_refs = frozenset(
+        m.ref for p in registry.plans.values() if p.key == plan_key
+        for m in p.models.values()
+    )
+    relevant_lanes = [
```
Fledgewing, line 58: **[nit] Docstring promises "the lane the session was reachable from"; code iterates all lanes naming the drained plan**

The `migrate` docstring (lines 36-44) says:

> "a session is migrated through the first lane that names a ref on the drained plan (the lane the session was reachable from before draining started)"

…but `relevant_lanes` (line 51) builds the full set of lanes that have *any* member on the drained plan, and the inner loop at line 60 tries them in `registry.lanes` iteration order. A ref shared by multiple lanes (e.g., one `claude-max/fable` appearing in a `forge` lane and a `tools` lane) lets a session that originated on lane A be re-leased onto a sibling reachable only via lane B. Both legs are reachable siblings, but the lane identity — and therefore the visible lane on the capacity board, the click-target on the portal — can change without warning.

Two ways to close the gap, pick whichever matches the intended contract:

1. Tighten the docstring to what the code actually does:

> "a session is migrated through any lane that names a ref on the drained plan; the first lane in registry iteration order whose `pick` succeeds wins."

2. Pin the migrator to the lane the session’s current ref was reachable from. `sessions_on_plan` only returns session ids, so we look up the lease first:

```python
sessions: list[tuple[str, str | None]] = []
for session in await slots.sessions_on_plan(plan_key):
    held = await slots.get_lease(session)
    sessions.append((session, held))

for session, held in sessions:
    placed = None
    # Prefer the lane the held ref was reachable from; fall back to any.
    candidate_lanes = []
    if held:
        candidate_lanes = [lane for lane in registry.lanes
                          if held in {m.ref for m in registry.lane_members(lane)}]
    candidate_lanes = candidate_lanes or relevant_lanes
    for lane in candidate_lanes:
        try:
            placed = await picker.pick(lane, session, exclude=drained_refs)
            break
        except Exception:
            continue
    ...
```

The tests currently don’t pin this either way — `test_drain_migrate_rewrites_n_leases_to_the_sibling_ref` only asserts the rewritten lease is on a non-drained plan, never that it’s on the same lane the session started in — so changing the contract will require a test, not just code.

Fledgewing:
Reviewed 5 findings at `16443651` — 1 blocker (picker comment vs apply.sh ordering + fragile reliance on affinity's exclude drop), 2 should-fix (drain-count grep, touch_lease race), 2 nit (dead `last_exc`, drain.py lane-iteration drift). See inline comments.

switchyard/slots.py
```diff
@@ -246,19 +314,73 @@ async def get_lease(self, session: str) -> str | None:
...
+                        plan_key: str | None = None) -> None:
+        # `ref` is the model ref (e.g. `claude-max/fable`); the plan key is
+        # the prefix before `/`. Callers in the picker pass it explicitly so
+        # the plan lookup is done once, but it is optional for tests that
+        # already have the ref and don't want to thread a registry through.
+        if plan_key is None:
+            plan_key = ref.split("/", 1)[0]
+        await self.redis.set(K_LEASE.format(session=session), ref, ex=ttl)
+        await self.redis.sadd(K_LEASE_PLAN.format(plan=plan_key), session)
+        await self.redis.expire(K_LEASE_PLAN.format(plan=plan_key), ttl)
```
Fledgewing: **[should-fix] `set_lease` has the same SET/SADD/EXPIRE race that cycle-2 just closed for `touch_lease` and `drop_lease`**

```python
async def set_lease(self, session: str, ref: str, ttl: int,
                    plan_key: str | None = None) -> None:
    if plan_key is None:
        plan_key = ref.split("/", 1)[0]
    await self.redis.set(K_LEASE.format(session=session), ref, ex=ttl)
    await self.redis.sadd(K_LEASE_PLAN.format(plan=plan_key), session)
    await self.redis.expire(K_LEASE_PLAN.format(plan=plan_key), ttl)
```

Cycle-2 closed the analogous race in `touch_lease` (EXPIRE then GET losing the SET TTL refresh) and in `drop_lease` (GET then DEL+SREM racing re-creates with the wrong plan key) by moving both into `_TOUCH_LEASE` / `_DROP_LEASE` Lua scripts where the read-and-write is one round-trip. `set_lease` is the third member of the same trio and is left untouched.

The residual race is between the `SADD` on line 326 and the `EXPIRE` on line 327 — a single round-trip. If Redis fails the EXPIRE (network drop, server error, client timeout) while the SADD succeeded, the SET has the new membership but no TTL. `SADD` does not refresh an existing TTL and does not set one on a fresh key, so the membership persists **forever** instead of healing at `lease_ttl_seconds`. Repeated migrations on that plan would then walk `sessions_on_plan(plan)`, see the stale session, hand it to `picker.pick(lane, session, exclude=drained_refs)` — pick succeeds on a sibling and `set_lease` writes a *new* SET entry on the new plan, but does **not** SREM from the old one, so the stale entry on the old plan survives every migrate pass.

Downstream effects:

- The migrator re-picks and re-prints the same session every drain — wasted work and noisy logs.
- The board’s `considered` list (and any future “sessions per plan” surface) reads a phantom for any plan whose `set_lease` ever partial-failed.

The fix is the same shape as cycle-2’s other two:

```lua
-- KEYS[1] lease key, KEYS[2] reverse-index key
-- ARGV[1] session, ARGV[2] ref (already plan-prefixed by caller), ARGV[3] ttl
local added = redis.call('SADD', KEYS[2], ARGV[1])
redis.call('SET', KEYS[1], ARGV[2], 'EX', ARGV[3])
redis.call('EXPIRE', KEYS[2], ARGV[3])
return added
```

…and the same one-script-atomicity reasoning applies: the SADD/EXPIRE pair cannot disagree on TTL, and there is no window where the SET exists without one. Add `if "SET_LEASE" in src:` in `FakeRedis.register_script` for parity with the other shadows, and a test in `test_drain.py` that asserts a SET’s TTL survives one heartbeat plus a deliberate half-failure simulation (today the fake doesn’t model per-key TTL on `self.sets`, so the test would need to add that — same pre-existing gap as the TOUCH_LEASE shadow comment already calls out).

Consistency argument: leaving `set_lease` as the one non-atomic lease method, while every other lease path is Lua-scripted, is exactly the kind of asymmetry a future maintainer will trip over. Fix it now, and the three-method lease lifecycle has one shape.

Fledgewing:
Reviewed 1 finding at `517d589` — cycle-1's 5 findings all addressed (picker comment truth, apply.sh awk parser, atomic Lua for touch_lease + drop_lease, dead `last_exc` removed, drain.py docstring matches code); the new `test_affinity_drops_lease_when_held_plan_is_draining` covers the new affinity drain-check and all 5 drain tests + the 19-test offline suite pass. The one cycle-2 finding is the third member of the lease trio: `set_lease` still has the SET/SADD/EXPIRE race shape the diff closed for `touch_lease` and `drop_lease` — see inline comment at switchyard/slots.py:327.

PR #91, opened 2026-09-23
Implement issue #75: add per-lane `image_routing` (`always | only_first | off`, default `off`) and per-model `supports_images` flag

### LINKED ISSUE #75: Route requests with images to an image-capable provider (per-lane mode)
### Problem

Some models in a lane don't accept image inputs (or accept them but won't process them usefully — e.g. a text-tuned model on a vendor that does support vision on a different model). Today nothing in the picker distinguishes "this request has an image" from a plain text turn, so a session that started on a text-only model will keep landing on it for every subsequent turn, including turns that carry images.

### Proposal

Add a per-lane `image_routing` setting with three modes. The picker uses it to filter and to decide what to do with an existing session lease; it does **not** bypass or reorder the lane's own routing mechanism — the existing body walk (round_robin / weighted / perishable / lowest_utilization / flat) is unchanged, this only narrows the candidate set and adjusts the affinity check.

### Modes

| Mode | First request has images | First request is text, later request adds images |
| --- | --- | --- |
| `always` | Route to an image-capable model | Move the session to an image-capable model (drop the old lease, walk the body for an image-capable candidate, lease onto it) |
| `only_first` | Route to an image-capable model | Stay on the existing model; whatever it does with the image is its problem |
| `off` (default) | Existing lane routing, regardless of images | Existing lane routing, regardless of images |

### Config shape

Per-lane, alongside the existing `strategy` / `tail` knobs:

```yaml
lanes:
  default:
    image_routing: always   # always | only_first | off
```

### Detection

A request "has an image" if any message in the request payload contains an `image*` content block (URL, base64, file_id, etc., per the OpenAI / Anthropic content shapes we already accept). Detection happens in `hooks.py` next to the existing `needs_tools` derivation (`switchyard/hooks.py:274`), and is passed into `picker.pick(...)` the same way.

### Implementation sketch

Mirrors the existing `needs_tools` / `can_use_tools` pattern — `picker.py:171-180` is the closest precedent — but with the capability flag on `Model`, not `Plan`. The reasoning: image support is a property of the model (the model has vision weights), not of the plan tier. Two models on the same plan commonly differ — a vision model and a text-only model behind one API key — and a plan-level flag would falsely grant vision to both. Tool support is genuinely plan-gated (subscription tier / endpoint access — see `Plan.can_use_tools` at `models.py:268`), so the two capabilities should not share a location just because they're both "capability flags".

1. Add `Model.supports_images: bool` to `switchyard/models.py` (next to `enabled` / `max_parallel` / `context_window` around `models.py:92-100`), sourced from a new `supports_images:` key per model in `config/plans.yaml`. **Default `false`** so a missing field never silently grants vision. For self-hosted / Ollama / custom providers with no central registry, operators set it explicitly.

2. **Enrichment from LiteLLM** at config-load time (optional, in `gen_litellm.py` or wherever `Registry` is built): LiteLLM ships a `model_cost` dict that flags known-vision models. Where `supports_images` is unset, pre-fill `true` from that metadata. Operator-set values still win; the enrichment only fills the unset cases. This saves the operator from repeating what's already known for the big vendors and adds no runtime probe.

3. In `picker.py`:
   - `_members` filters out non-image models when `ctx.needs_images and mode == "always"` (the existing `if needs_tools:` block at `picker.py:177-180` is the template, just swap `self.registry.plan_of(m).can_use_tools` for `m.supports_images`).
   - `_visit_ref` re-checks per-leaf (same template as `ctx.needs_tools and not plan.can_use_tools` at `picker.py:237`, swap to `ctx.needs_images and not model.supports_images`).
   - `_affinity` drops the lease and falls through to the body walk when `mode == "always"` and the held model can't serve images (mirrors the existing `held in ctx.exclude` / `not plan.can_use_tools` spills at `picker.py:616-627`, swap to a `not model.supports_images` check). For `only_first`, affinity behaves as today and the held model either processes the image or returns whatever error it returns.

4. Surface the filter on the capacity board the way `cap_reason: "model limit N"` already does, so operators can see "this candidate was filtered for image support" rather than "this candidate was unavailable".

5. Add a `image_routing` field to `LaneConfig` in `switchyard/models.py`, defaulting to `"off"`, so existing configs are bit-for-bit unchanged.

### Non-goals

- This is **not** an image-content fallback (no rewriting the payload to text, no OCR-on-the-gateway). If no image-capable model in the lane has capacity, the request gets the existing `LaneSaturated` 429 — same as a lane with no live members today.
- This is **not** image *generation*. It is image *input* / vision.
- This does **not** introduce a new lane-level routing strategy; it is a filter applied within the strategy the lane already has.

### Resolved

- **Per-model, not per-plan.** Image support is a model capability (vision weights), not a plan-tier gate. Plan-level would falsely grant vision to a text-only model sharing the plan. Distinct from `supports_tools`, which is plan-gated.
- **Defaults in `config/plans.example.yaml`.** Both CLI sidecars the operator's config ships with already accept image input natively:
  - `claude-max-sidecar` → Claude Code accepts pasted images via `Ctrl+V` / `Cmd+V` / `Alt+V` from the clipboard ([interactive-mode docs](https://code.claude.com/docs/en/interactive-mode), "Paste image from clipboard").
  - `codex-sidecar` → Codex CLI accepts image files via `codex -i FILE` and pasted/dragged images in the interactive composer ([image-inputs docs](https://developers.openai.com/codex/image-inputs)).

  So models under the `claude-max` plan and the `codex-sidecar` plan can ship with `supports_images: true` by default; everything else stays `false` until the operator confirms. **Caveat:** the sidecar implementations still have to actually forward image content through to the underlying CLI — assumed true here (it has to be for the existing flow to work today), worth confirming in each sidecar before flipping the default on. This is a routing decision, not a content-transport guarantee.

### Still open

- Should `always` also apply to the first request when the operator configured `only_first` and a later turn upgrades the session — i.e. promote the session's mode? (Default proposal: no, the mode the operator wrote is the mode that runs.)
- Naming: `image_routing` (lane config) vs `supports_images` (model field) — current proposal. `vision_routing` / `multimodal_routing` considered and rejected as longer words for the same thing.

### Related

- #30 (closed) — `cli_bridge` / `mcp_bridge` silently drop image blocks during `flatten()`. This proposal does not depend on #30, but if #30's fix ever lands, the picker filter will start seeing image-bearing requests it never saw before.

Fledgewing:
Fixes #75.

## Summary

Add a per-lane `image_routing` (`always | only_first | off`, default `off`)
and a per-model `supports_images` flag. Under `image_routing: always`, an
image-bearing request filters non-image-capable members and drops an
existing session lease held on a text-only model so the next image turn
re-places the session on a member that can serve it. `only_first` keeps
today's sticky affinity (a forward-compatible opt-in for image-only
first-turn routing); `off` (default) is bit-for-bit legacy routing — an
image-bearing request behaves exactly like a text-only one. The lane's
own strategy (`round_robin` / `weighted` / `perishable` /
`lowest_utilization` / `fill`) is untouched: `image_routing` only narrows
the candidate set and adjusts the affinity check, never bypasses the
strategy, and when every image-capable member is out of capacity the
request gets the usual `LaneSaturated` 429.

## Changes

- **`switchyard/models.py`** — Add `supports_images: bool | None` to
  `Model` and `image_routing: str = "off"` to `Lane`. `_parse_models`
  reads `supports_images` as a tri-state so operator-set `True` / `False`
  is never overwritten by the loader. `load()` reads
  `litellm.model_cost` via a lazy-import helper
  (`litellm_known_vision_models`) to pre-fill `True` for any model
  whose key or `model:` string matches a known vision entry — but
  operator values always win. `image_routing` is validated against
  `always | only_first | off` with a loud `ValueError` on typos.
- **`switchyard/picker.py`** — Plumb `needs_images` and `image_mode`
  through `_VisitCtx`. `pick()` resolves `image_mode` from the lane.
  `_members()` filters non-image-capable members under `always`;
  `_visit_ref()` re-checks per-leaf and appends a `"(no image support)"`
  skip note; `_affinity()` drops the lease under `always` so the body
  walk re-places the session. `LaneSaturated` detail gains an images
  case mirroring the existing tool-capable wording.
- **`switchyard/hooks.py`** — Add `_has_image_block()` (Anthropic
  `image`, OpenAI Chat `image_url`, OpenAI Responses `input_image`).
  Derive `needs_images` from the message list, pass it into `pick()`,
  stamp `needs_images` into the metadata ctx, and append `" images"` to
  the routing log line. `pick_direct` is untouched — a caller naming a
  deployment asked for exactly that model.
- **`config/plans.example.yaml`** — Document `supports_images: true` on
  the two vision-capable plans (claude-max subscription: opus, fable,
  sonnet; openai codex sidecar: sol, terra, luna, astra). Add the
  `image_routing` key with explanatory comment to every lane (default
  `off`). The live `config/plans.yaml` is gitignored and operator-managed.

## Tests

11 new picker tests + 4 new hook tests, all offline (loopback only,
no real providers). All tests green:

- `test_image_routing_default_off_leaves_routing_unchanged_for_image_turns`
- `test_members_filters_to_image_capable_under_always`
- `test_image_routing_always_lets_a_lane_with_no_text_only_pick`
- `test_text_only_lane_under_always_raises_lane_saturated_with_skip_reason`
- `test_visit_ref_per_leaf_skips_text_only_under_always_with_explicit_reason`
- `test_affinity_spills_when_image_request_hits_text_only_lease_under_always`
- `test_only_first_keeps_sticky_pick_on_text_only_for_image_request`
- `test_supports_images_parser_handles_true_false_and_missing`
- `test_lane_image_routing_default_is_off`
- `test_lane_image_routing_validates_mode`
- `test_litellm_known_vision_returns_a_set_with_vision_models`
- `test_pre_call_hook_derives_needs_images_from_anthropic_image_block`
- `test_pre_call_hook_derives_needs_images_from_openai_image_url_block`
- `test_pre_call_hook_derives_needs_images_from_openai_input_image_block`
- `test_pre_call_hook_leaves_needs_images_false_for_plain_text`

The picker tests cover: `_members` filtering under `always`, an
all-text-only lane refusing with the `"(no image support)"` skip note,
affinity spill (old lease dropped, new lease on image-capable model),
`only_first` keeping the sticky pick on a text-only model, `off` being
bit-for-bit identical to the legacy path, parser tri-state (True /
False / None), loader-default `off`, and a loud `ValueError` on bogus
mode values. The hook tests cover all three protocol image shapes
(Anthropic / OpenAI Chat / OpenAI Responses) end-to-end through the
pre-call hook, plus the negative plain-text case.

## Verification

```
python3 tests/test_routing.py && \
python3 tests/test_messages_hooks.py && \
python3 tests/test_hot_reload.py && \
for t in tests/test_*.py; do python3 "$t" >/dev/null || echo "FAIL $t"; done && \
```

All commands green; `tests/test_routing.py` reports the 11 new
image-routing tests passing; `tests/test_messages_hooks.py` reports the
4 new image-detection tests passing; `render_preview.py` reports
`10 plans, 19 models, 6 lanes, 18 quota windows, 4 probes`.


switchyard/picker.py
```diff
@@ -169,15 +175,29 @@ async def is_spent(self, plan) -> bool:
...
-        plan cannot serve this request at all."""
+        plan cannot serve this request at all.
+
+        Image filtering applies only when the lane's `image_routing` is set to
+        "always" -- "off" and "only_first" leave the member set untouched so
+        the lane's own routing strategy stays load-bearing. The picker passes
+        `needs_images` but ALSO reads the lane's mode internally because the
+        call site that builds `ctx` already knows it; this filter applies the
+        same rule at the body-build step so a request whose body is wholly
+        text-only never silently widens the candidate set under "always".
```
Fledgewing: **[should-fix] Picker._members docstring: "never silently widens" reads backwards**

This filter NARROWS the candidate set under `image_routing: always` — it
drops members whose `supports_images` is falsy. It never widens anything.
The docstring reads:

> ...so a request whose body is wholly text-only never silently widens
> the candidate set under "always".

A reader has to re-parse "widens" as "is unaffected (neither widened nor
narrowed)" to make sense of it, which is the opposite of what the word
means.

The intended point is that the filter is gated on `needs_images=True`,
so a text-only request (`needs_images=False`) under `always` sees the
same candidate set it would have before this PR — i.e. the picker does
not retroactively treat a text request as image-bearing and shrink the
set behind the operator's back. A clearer phrasing:

> The filter is gated on `needs_images=True`, so a text-only request
> under `always` sees the unchanged member set — image routing only
> kicks in when the request itself carries image blocks.

Also: the middle sentence ("The picker passes `needs_images` but ALSO
reads the lane's mode internally because the call site that builds
`ctx` already knows it") is hard to parse. `_members` is called from
`pick()` with `needs_images`; `pick()` has already resolved
`image_mode` into `ctx`. `_members` reads `self.registry.lanes[...].image_routing`
directly instead of accepting ctx, so the "call site knows it" claim
isn't really explaining a design choice — it's documenting an
inconsistency between where `needs_images` flows through and where the
mode is read. Either thread `ctx` (or at least `image_mode`) into
`_members` so the two halves agree, or rewrite the sentence to say
"`_members` re-resolves the lane's `image_routing` rather than threading
`ctx` through, so an off-mode lane and an unset-mode lane behave
identically."

tests/test_routing.py
```diff
@@ -4281,6 +4281,488 @@ async def go():
...
+    try:
+        body = open(os.environ["SWITCHYARD_PLANS"]).read()
+        # Drop the image_routing key from one lane and confirm `off`.
+        # The shipped fixture sets it for every lane; we strip from apex.
+        import re as _re
+        stripped = _re.sub(
+            r"    # image_routing only.*?image_routing:\s*\"off\".*?\n",
+            "", body, count=1, flags=_re.DOTALL,
+        )
+        # The comment block above also has `image_routing` mentioned in
```
Fledgewing: **[nit] test_lane_image_routing_default_is_off: stale comment describes an action that is not performed**

```python
stripped = _re.sub(
    r"    # image_routing only.*?image_routing:\s*\"off\".*?\n",
    "", body, count=1, flags=_re.DOTALL,
)
# The comment block above also has `image_routing` mentioned in
# its prose, so re-add a stripped version that's id-mismatched.
with os.fdopen(fd, "w") as fh:
    fh.write(stripped)
```

The inline comment on lines 4697-4698 says "re-add a stripped version that's id-mismatched" — but the code that follows it just writes `stripped` verbatim. There is no re-add step. It looks like a leftover from a refactor: the comment is describing an alternative approach that was considered and abandoned, not the approach the test is actually taking.

The test still works because the `count=1` regex strips exactly one lane's worth of comment+field (apex's), and `image_routing: "off"` is the dataclass default, so loading the stripped YAML still yields `apex.image_routing == "off"`. So this is not a correctness bug, just dead documentation.

Either drop the two-line comment or rewrite it to describe what the code is actually doing (e.g., "regex strips both lines — the multi-line comment block plus the `image_routing: "off"` field — for the first lane only").

config/plans.example.yaml
```diff
@@ -669,15 +669,23 @@ plans:
       opus:
         model: openai/claude-opus-5
         max_parallel: 2
+        # Sidecar accepts image input natively. Set true explicitly so the
+        # picker routes image-bearing requests here under image_routing:
+        # always; omit (defaults false) for text-only plans.
```
Fledgewing: **[should-fix] plans.example.yaml: "omit (defaults false)" is misleading — the actual default is unset and litellm may auto-promote it**

The `opus` model comment block reads:
```yaml
# Sidecar accepts image input natively. Set true explicitly so the
# picker routes image-bearing requests here under image_routing:
# always; omit (defaults false) for text-only plans.
supports_images: true
```

The "(defaults false)" advice is misleading in two ways:

1. The picker-level default is NOT `False`. `Model.supports_images` is a
   tri-state (`bool | None`). When the operator omits the key, the field is
   `None`, and `load()` only converts `None → True` via litellm pre-fill
   for models whose `model:` string matches a `litellm.model_cost` entry
   with `supports_vision=True`. From the picker's perspective `None` is
   treated as "no image support" (`not None is True`), but the underlying
   value is NOT the boolean `False` — it is `None`, and the
   `litellm_known_vision_models()` pre-fill can flip it to `True`
   behind the operator's back.

2. In this specific example, the comment would be confusing if an operator
   copied this stanza to a model whose `model:` string happens to match a
   litellm entry. For instance, if someone wrote `model: gpt-4o` and
   omitted `supports_images`, the loader would silently set it to `True`
   at `load()` time, and the picker would route image-bearing requests
   to it under `image_routing: always`. The operator who followed the
   "omit (defaults false)" advice would think they configured a
   text-only model.

The four mirror copies (claude-max fable/sonnet and openai sol/terra/luna/astra)
all repeat this same comment, so the misleading text appears 7 times in
the shipped example. Suggested rephrasing:

```yaml
# Sidecar accepts image input natively. Set true explicitly so the
# picker routes image-bearing requests here under image_routing:
# always; omit to leave it unset, which the picker treats as "no
# image support" — but `load()` may pre-fill True from
# `litellm.model_cost` if the `model:` string matches a vision
# entry, so set false explicitly to opt out.
```

That at least surfaces the litellm-pre-fill gotcha.

config/plans.example.yaml
```diff
@@ -929,13 +947,25 @@ lanes:
     # strategy: perishable
     order: [claude-max/fable, openai/astra]
     tail: [local-box/qwen]
+    # image_routing only narrows the candidate set (and adjusts affinity under
+    # `always`); it never bypasses the lane's own routing strategy. When every
+    # image-capable member is out of capacity, the request gets the usual
+    # LaneSaturated 429. Default off: an image-bearing request behaves
+    # bit-for-bit the same as a text-only one.
+    image_routing: "off"     # off | always | only_first (default off)
```
Fledgewing: **[nit] plans.example.yaml: every lane is `image_routing: "off"` — there is no worked example of the feature in action**

All six lanes in the shipped fixture are pinned to `image_routing: "off"`. The block comment correctly explains the default and the modes, but a fresh operator who reads this config and the issue (#75) has no in-repo worked example of a lane that actually exercises the new routing — i.e. a lane with `image_routing: "always"` and a member list that includes a mix of image-capable and text-only models, where the picker would have to filter.

The risk is that an operator who copies `plans.example.yaml` to start their own `plans.yaml` ends up with `off` everywhere by default, decides later they want image routing, flips one lane to `always`, and is surprised when text-only members stop being considered (because they assumed `always` was a synonym for `off`). The opposite end of the spectrum is also confusing: a casual reader of the diff sees "image_routing added" and assumes the example config now demonstrates it.

Two low-cost options that would close this gap:

1. Set one lane to `image_routing: "always"` to anchor the pattern. The
   `apex` lane already has `claude-max/fable` and `openai/astra` — both of
   which are image-capable in the example — so flipping apex to
   `"always"` would not break anything (it just changes the no-op behaviour
   of the new code path from "ignore the flag" to "filter to image-capable,
   which here happens to leave every member standing").

2. Add a commented-out worked example in the lane preamble, e.g.:
   ```yaml
   # Worked example for image routing:
   #   vision-demo:
   #     label: "Vision"
   #     order: [claude-max/opus, openai/astra]    # both image-capable
   #     tail: [local-box/qwen]
   #     image_routing: "always"  # image-bearing requests land here
   ```

Either makes the new feature discoverable from the example alone, without
having to read the picker source or the issue body.

Fledgewing:
Reviewed 4 findings: 2 should-fix, 2 nit — see inline comments at 75a2741.

PR #101, opened 2026-09-23
Fix issue #89 (Fledgewing/SwitchYard): the mcp_bridge sidecar reports the inner

### LINKED ISSUE #89: mcp_bridge reports cumulative token usage, not conversation size — Claude Code auto-compacts after every tool-using reply
## Symptom

Claude Code reads each reply's `usage` as "how full is my context" and triggers an auto-compact once that number passes the window.

Through SwitchYard's `claude-pool` (mcp_bridge), the value Claude Code sees is wildly larger than the actual conversation. As a result, **every reply that used tools triggers a compaction**. Observed live — one tab compacted 6 times in an hour.

Plain text replies (no tools) are not affected. Tool-using replies all do.

## What Claude Code sees

For a single caller turn that spans many tool rounds, the inner CLI's final `usage` is the sum of every model call the inner session made. Twenty tool rounds over a ~100k context come back reported as ~2M input tokens, and Claude Code compacts.

Measured on real sessions through the gateway:

| Real conversation size | Reported on the final turn | Result |
|---|---|---|
| ~95k | 1,676,761 | auto-compact |
| ~170k | 9,947,000 | auto-compact |
| 4-turn test, ~17k | 86,763 | 5x too big |

Intermediate tool-call turns report 0 (or omit the field entirely), so the caller sees a meter that jumps from 0 to "way past the window" with no warning.

## Why this is a SwitchYard problem, not a Claude Code problem

The mcp_bridge path is one inner CLI run that carries the whole tool-call loop inside a single outer `result`. The `usage` it surfaces is whatever the inner CLI's last call reported. For a Claude-protocol caller that reads `usage` to decide on compaction, that's the wrong number — it's the run's spend, not the conversation's size.

Whatever the right fix is, it has to make the value Claude Code sees match the actual size of the conversation it is keeping, not the token spend of the multi-step run that produced the reply.

## Repro

1. Pick any `claude-pool` tab (an mcp_bridge plan, e.g. `sy.claude-pool` / claude-max or another Claude-protocol provider).
2. Send a prompt that requires tool use (e.g. "find the file with X, read it, summarize").
3. Inspect the final reply's `usage` block as the caller sees it.
4. Compare to the actual conversation size by any honest measure (tokenizer, manual count of messages and lengths).

Observed mismatch: 5x – 50x depending on tool round count.

## Scope

- All `claude-pool` / mcp_bridge lanes serving Claude Code or any other caller using Claude Code's compaction heuristic.
- Tool-using replies only — plain replies go through fine.
- Affects every Claude-protocol caller that reads `usage` to gate compaction, not just the literal "Claude Code" client.

## Evidence

Live observation: one tab, 6 compactions in an hour, each one followed by a tool-using reply. The 4-turn test above was run end-to-end through the gateway and the 86,763 reading is what the caller's `usage` block showed.

The PR that accompanied this report (`#88`) proposes one way to address it; I'm filing it as a bug so the team can pick the approach.

Fledgewing:
## What goes wrong

An mcp_bridge session is one inner CLI run that spans every tool round of a caller turn. The CLI's final `usage` is the **sum of all its model calls**, not the size of the conversation the caller is keeping. Twenty tool rounds over a ~100k context come back as ~2M input tokens, and the caller auto-compacts after every tool-using answer.

Measured on real sessions through the gateway:

| Real conversation | Reported on the final turn | Result        |
|-------------------|----------------------------|---------------|
| ~95k              | 1,676,761                  | auto-compact  |
| ~170k             | 9,947,000                  | auto-compact  |
| 4-turn test, ~17k | 86,763                     | 5x too big    |

Tool-call turns reported `0`, so the meter sat at zero and then jumped past the window.

## The fix

- `sidecars/mcp_bridge/server.py`
  - `last_call_usage(session)` reads the inner CLI's own transcript (the per-call `usage` it writes for every model call) and returns the most recent assistant call's numbers. That is the conversation's real size.
  - `render_turn` reports that on tool-call turns and on the final turn via `_with_context_usage`, which folds it through `cli_bridge.to_openai` so cache reads/writes keep their shape.
  - The run's actual spend is carried alongside as `switchyard_billed_prompt_tokens` / `switchyard_billed_completion_tokens` on both turn types — the OpenAI-shaped sum the CLI's payload carried on the final turn, an empty dict mid-loop.
  - Only the `claude` provider has a transcript; without one the old usage shape stays as today (verified by test).
- `switchyard/hooks.py`
  - `_prompt_completion_tokens` books the billed figures when stamped, on both dict and attribute-shaped usage objects, so the gateway ledger sees the run's real spend rather than the conversation-context number. All other paths stay byte-for-byte unchanged.
  - On the LiteLLM `/v1/messages` translation path the extra keys may be dropped; per-turn prompt sizes there add up to the same total, so the ledger stays correct either way.

## Tests

`python3 tests/test_mcp_bridge.py && for t in tests/test_*.py; do python3 "$t" >/dev/null || echo "FAIL $t"; done` — full suite green.

Six new tests cover:

- `test_final_turn_reports_last_call_context_size_and_bills_the_sum` — final turn with a 20-call summed payload: `usage.prompt_tokens` carries the LAST call's size, `switchyard_billed_*` carry the sum, and the cache breakdown (`prompt_tokens_details.cached_tokens`, top-level `cache_creation_input_tokens`) rides alongside as on the text path.
- `test_no_inner_transcript_keeps_old_usage_and_bills_mirror_it` — missing project dir and empty project dir both keep today's usage unchanged, with billed keys mirroring the kept figures.
- `test_prompt_completion_tokens_books_billed_over_reported` — usage dict with context-size prompt + billed keys books the billed amounts.
- `test_prompt_completion_tokens_plain_dict_unchanged` — plain usage dict still goes through the existing extraction.
- `test_prompt_completion_tokens_billed_attribute_shape` — same override fires on the pydantic / attribute-shaped usage object path.
- `test_prompt_completion_tokens_billed_completion_optional` — billed-completion key allowed to be absent / `None`; override still fires with completion defaulting to 0.

Fixes #89


sidecars/mcp_bridge/server.py
```diff
@@ -1266,22 +1267,123 @@ async def internal_tools_call(request: Request):
...
             "created": int(time.time()), "model": requested_model or session.model,
             "choices": [{"index": 0, "message": message, "finish_reason": "tool_calls"}],
             # Token accounting for the CLI-backed path is already approximate
             # (see cli_bridge); mid-loop it is not available at all.
             "usage": {"prompt_tokens": 0, "completion_tokens": 0, "total_tokens": 0},
         }
+        # No usage has been booked yet on a mid-loop tool_calls turn, but
+        # the caller can still want a real context meter — the CLI wrote
+        # one to its transcript before exiting this turn.
+        return _with_context_usage(response, session, {})
```
Fledgewing, line 1380: **[should-fix] Mid-loop tool_calls turn path is not covered by any test**

`_with_context_usage(response, session, {})` is reached on every tool_calls
turn (mid-loop), but the two new mcp_bridge tests only exercise the FINAL
turn path:

* `test_final_turn_reports_last_call_context_size_and_bills_the_sum` —
  final turn with a fully populated transcript.
* `test_no_inner_transcript_keeps_old_usage_and_bills_mirror_it` — final
  turn with NO transcript (both flavours: missing project dir, empty
  project dir).

Neither test lays down a transcript and calls `render_turn` with a
`result["type"] == "tool_calls"` dict. So the branch where
`last_call_usage` returns non-None on a tool_calls turn — i.e. exactly the
mid-loop case the PR description leads with ("Tool-call turns reported 0,
so the meter sat at zero and then jumped past the window") — has no
direct test coverage. The behaviour was inferred from the final-turn test,
but `_with_context_usage` mutates `usage` in two different ways for the
two turn types (replacement from `cli_bridge.to_openai` for tool_calls,
folding billed_* onto the existing usage for final), so they should be
pinned separately.

Suggested addition, mirrored on `_write_inner_transcript`:

    result = {"type": "tool_calls", "calls": [ParkedCall(id="c1",
                name="get_weather", arguments={"city": "Oslo"}, ...)]}
    response = server.render_turn(session, result, session.model)
    assert response["usage"]["prompt_tokens"] == expected_prompt
    assert response["usage"]["switchyard_billed_prompt_tokens"] == 0  # billed empty on mid-loop

tests/test_mcp_bridge.py
```diff
@@ -3230,6 +3232,176 @@ async def go_codex():
...
+            "input_tokens": 500,
+            "output_tokens": 200,
+            "cache_read_input_tokens": 0,
+            "cache_creation_input_tokens": 0,
+        }
+        result = {"type": "final", "payload": {"result": "answer", "usage": summed}}
+        response = server.render_turn(session, result, session.model)
+        usage = response["usage"]
+        # Today would be: prompt_tokens=500, completion_tokens=200,
+        # total_tokens=700 -- the to_openai shape with no Anthropic cache
```
Fledgewing, line 3490: **[nit] Test comment is misleading about the Anthropic-shape branch**

```python
summed = {
    "input_tokens": 500,
    "output_tokens": 200,
    "cache_read_input_tokens": 0,        # key PRESENT
    "cache_creation_input_tokens": 0,    # key PRESENT
}
...
# Today would be: prompt_tokens=500, completion_tokens=200,
# total_tokens=700 -- the to_openai shape with no Anthropic cache
# keys, so prompt_tokens is just input_tokens. That stays put.
```

The comment claims "no Anthropic cache keys", but `cache_read_input_tokens`
and `cache_creation_input_tokens` are both in the dict — they just happen
to be zero. `cli_bridge.to_openai` keys its shape detection on `in usage`,
not on a non-zero value:

```python
anthropic_shape = ("cache_read_input_tokens" in usage
                   or "cache_creation_input_tokens" in usage)
```

so this test is actually exercising the Anthropic-shaped branch (with
both cache counts folding into `prompt_tokens` as `500 + 0 + 0 = 500`).
The assertion still passes because adding zeros is a no-op, but the
comment misrepresents the path being tested. Reword to something like
"with cache keys at zero, so the Anthropic-shape fold leaves
prompt_tokens = input_tokens = 500" — or add a no-cache-keys flavour to
pin the OpenAI-shape branch separately.

Fledgewing:
Reviewed 2 findings: 1 should-fix, 1 nit — see inline comments at 7733e57.

Highlights:
* The mid-loop tool_calls turn path through `_with_context_usage(response, session, {})` is not exercised by any of the six new tests; both mcp_bridge tests only cover the FINAL turn path.
* A test comment in `test_no_inner_transcript_keeps_old_usage_and_bills_mirror_it` misrepresents the Anthropic-shape detection as absent when the keys are present at zero.

All 39 test_usage + 62 mcp-bridge tests pass; full offline suite (`for t in tests/test_*.py`) green. No blockers, no security issues, behaviour matches the PR description on the paths covered by tests.

